PR #9, opened 2021-07-12
Add boolean support

Kommandat:
Support for 8-bit (1 byte) booleans via a new `q_compress::types::boolean8::B8DataType` type

Some references:
- https://users.rust-lang.org/t/solved-what-is-byte-value-for-boolean/47347
- https://doc.rust-lang.org/std/primitive.bool.html

### DIFF
--- a/example/generate_randoms.py
+++ b/example/generate_randoms.py
@@ -29,6 +29,21 @@ def write_i64(arr, name):
   pq.write_table(table, f'data/parquet/i64_{name}.parquet', compression='NONE')
   pq.write_table(table, f'data/snappy_parquet/i64_{name}.snappy.parquet', compression='snappy')
 
+def write_bool8(arr, name):
+  if arr.dtype != np.int8:
+    floored = np.floor(arr).astype(np.int8)
+  else:
+    floored = arr
+  ints = [str(x) for x in floored]
+  joined = '\n'.join(ints)
+  with open(f'data/txt/bool8_{name}.txt', 'w') as f:
+    f.write(joined)
+  with open(f'data/binary/bool8_{name}.bin', 'wb') as f:
+    f.write(floored.tobytes())
+  table = pa.Table.from_pydict({'nums': floored})
+  pq.write_table(table, f'data/parquet/bool8_{name}.parquet', compression='NONE')
+  pq.write_table(table, f'data/snappy_parquet/bool8_{name}.snappy.parquet', compression='snappy')
+
 def write_f64(arr, name):
   arr = arr.astype(np.float64)
   floats = [str(x) for x in arr]
@@ -89,3 +104,5 @@ edge_case_floats[p < 0.3] = np.nan
 edge_case_floats[p < 0.2] = -np.nan  # yes, it is different
 edge_case_floats[p < 0.1] = np.NINF
 write_f64(edge_case_floats, 'edge_cases')
+
+write_bool8(np.random.randint(2, size=n), 'random')
--- a/src/types/boolean8.rs
+++ b/src/types/boolean8.rs
@@ -1,0 +1,42 @@
+use std::cmp::Ordering;
+use std::convert::TryInto;
+
+use crate::compressor::Compressor;
+use crate::decompressor::Decompressor;
+use crate::types::{DataType, NumberLike};
+
+impl NumberLike for bool {
+  fn num_eq(&self, other: &Self) -> bool {    
+    self.eq(other)
+  }
+
+  fn num_cmp(&self, other: &Self) -> Ordering {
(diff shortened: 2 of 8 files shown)


mwlon:
We really want single-bit booleans. We could already cast booleans to `u32` and use an existing compressor, but we should make it convenient to work with boolean data

example/generate_randoms.py
```diff
@@ -29,6 +29,21 @@ def write_i64(arr, name):
...
+def write_bool8(arr, name):
+  if arr.dtype != np.int8:
+    floored = np.floor(arr).astype(np.int8)
+  else:
+    floored = arr
+  ints = [str(x) for x in floored]
+  joined = '\n'.join(ints)
+  with open(f'data/txt/bool8_{name}.txt', 'w') as f:
+    f.write(joined)
+  with open(f'data/binary/bool8_{name}.bin', 'wb') as f:
```
mwlon, line 41: How about we instead pack the booleans into bytes? The raw data should in principle be 1/8 as big. But this is just for the example, so not totally essential

src/types/boolean8.rs
```diff
@@ -0,0 +1,42 @@
+use std::cmp::Ordering;
+use std::convert::TryInto;
+
+use crate::compressor::Compressor;
+use crate::decompressor::Decompressor;
+use crate::types::{DataType, NumberLike};
+
+impl NumberLike for u8 {
```
mwlon, line 8: Let's really make this for `bool` instead of `u8`. `bytes_from` and `from_bytes` can still use a whole byte though

Kommandat:
> We really want single-bit booleans. We could already cast booleans to `u32` and use an existing compressor, but we should make it convenient to work with boolean data

Got it. Should we also add support for `u8` then in case someone has booleans stored as 1 byte? Looking at most languages, this seems to be the most common option. This is the case for Rust — https://users.rust-lang.org/t/solved-what-is-byte-value-for-boolean/47347.

example/generate_randoms.py
Kommandat: Correct me if I'm wrong, but I was thinking the booleans are already packed into 8-bit bytes because of this line `floored = np.floor(arr).astype(np.int8)`

mwlon:
@Kommandat we'll be compressing whatever the user in memory, which is the `bool` type. I don't think we need a specific `u8` data type because the user can cast to `u32` and back trivially. The reason I want to do `bool` is that it's a very common type, and people might not think to cast it to `u32` or use this library for it otherwise.

src/types/boolean8.rs
```diff
@@ -0,0 +1,42 @@
...
+  fn num_eq(&self, other: &Self) -> bool {    
+    self.eq(other)
+  }
+
+  fn num_cmp(&self, other: &Self) -> Ordering {
+    self.cmp(other)
+  }
+}
+
+pub struct B8DataType {}
```
mwlon, line 18: Let's just call it BoolDataType, yeah?

src/types/boolean8.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl DataType<bool> for B8DataType {
+  const HEADER_BYTE: u8 = 4;
```
mwlon, line 21: Oh yeah, this byte just needs to be something unique. `u32` currently uses 4, so we should set this to something untaken, like 7. This just makes sure that when you read in `q_compressed` data you won't get mixed up about which data type it is

src/types/boolean8.rs
```diff
@@ -0,0 +1,42 @@
...
+  fn offset_diff(upper: bool, lower: bool) -> u64 {
+    (upper as u64) - (lower as u64)
+  }
+
+  fn add_offset(lower: bool, off: u64) -> bool {
+    (lower as u64 + off) != 0
+  }
+
+  fn bytes_from(value: bool) -> Vec<u8> {
+    (value as u8).to_be_bytes().to_vec()
```
mwlon, line 33: Another way to do this would be `vec![value as u8]`

src/types/boolean8.rs
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  fn from_bytes(bytes: Vec<u8>) -> bool {
+    u8::from_be_bytes(bytes.try_into().unwrap()) != 0
```
mwlon, line 37: Another way to do this would be `bytes[0] != 0`

example/generate_randoms.py
mwlon: Oh what I mean is that we can stuff 8 booleans into a single byte when storing on disk in a binary format. Not necessary to do this, and I can see an argument for leaving it as is.

PR #23, opened 2022-02-12
Rework BitReader to use byte slices

mcuelenaere:
Currently BitReader requires callers to give ownership of the data. Since it
actually never needs to have ownership, rework this so that BitReader instead
stores a reference to a byte slice. This allows callers not having to do unnecessary
copies of the raw data.

Note that this changes the public API (I haven't updated the changelog accordingly).

### DIFF
--- a/examples/fast_seeking.rs
+++ b/examples/fast_seeking.rs
@@ -21,7 +21,8 @@ fn main() {
 
   // now read back only the metadata
   let start_t = Instant::now();
-  let mut reader = BitReader::from(writer.pop());
+  let bytes = writer.pop();
+  let mut reader = BitReader::from(&bytes);
   let decompressor = Decompressor::<f64>::default();
   let flags = decompressor.header(&mut reader).expect("flags");
   let mut metadatas = Vec::new();
--- a/examples/primary.rs
+++ b/examples/primary.rs
@@ -28,7 +28,7 @@ trait DtypeHandler<T: 'static> where T: NumberLike {
       .simple_compress(&nums)
   }
 
-  fn decompress(bytes: Vec<u8>) -> Vec<T> {
+  fn decompress(bytes: &[u8]) -> Vec<T> {
     Decompressor::<T>::default()
       .simple_decompress(bytes)
       .expect("could not decompress")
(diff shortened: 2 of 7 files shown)


mwlon:
@mcuelenaere Thanks for the PR! I agree this is a good improvement for 0.7. I double-checked that the tests pass and performance is unharmed.

But could you please change the `simple_decompress` function to take a slice as well? I'd also be happy with accepting `AsRef<[u8]>`, but a regular old slice probably causes the least confusion for most users.

mcuelenaere:
@mwlon changed!

mwlon:
@mcuelenaere congrats! Out of curiosity, what are you using q_compress for?

mcuelenaere:
> @mcuelenaere congrats! Out of curiosity, what are you using q_compress for?

At Delta we have an autocompleter microservice that runs in a constrained environment (hard limit on CPU time and memory usage). Most of the code is proprietary, but a small part of it is open sourced at https://github.com/mcuelenaere/porigon.

I'm currently prototyping a change where the document IDs (which are fully kept in-memory), that are linked to a key of the FST (array of `uint64`s), are being compressed with q_compress (which has great compression rate!).

I was also contemplating changing the decompression interface of q_compress to be iterator based (to reduce the memory usage by not allocating `Vec<u64>`s), but so far it looks like that is premature/over optimization.

mwlon:
@mcuelenaere I've also thought about supporting decompression as an iterator or something similar. It requires a somewhat bigger change (probably surfacing `ChunkDecompressor` in the public API), but we could probably make it happen soon if people are interested.

mcuelenaere:
I justed tested the impact of this PR on query speed and noticed 5x to 10x improvements in request latency!

Examples:

| before | after      |
| ------ | -------- |
| 1.1s     | ~200ms |
| 900ms | 180ms  |
| 192ms | 21ms     |

Is it possible however that you merged this PR **after** you released 0.6.1? Are you planning to release this in a future release? (0.7?)

mwlon:
@mcuelenaere Yes, 0.6.1 was out before I merged this. I'll release this in 0.7.

Which versions were you comparing? Is the bottleneck of your request a q_compress decompression?

PR #29, opened 2022-02-17
changelog and version

mwlon:


mwlon:
@mcuelenaere your changes are now released, plus there's an API to decompress one batch at a time.

PR #32, opened 2022-02-19
added 16-bit and i128 support

### LINKED ISSUE #31: Support remaining integer types
Need `i16` support for my audio compression application. Might as well add `u16` and `i128` while there. Pull request incoming.

BartMassey:
This adds the remaining standard integer types as `NumberLike`. Closes #31.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -627,7 +627,7 @@ dependencies = [
 
 [[package]]
 name = "q_compress"
-version = "0.7.0"
+version = "0.7.1"
 dependencies = [
  "rand",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,4 @@
-[package]
-name = "q_compress"
-version = "0.7.1"
-edition = "2018"
+[workspace]
 
 members = [
   "q_compress",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,7 @@
-[workspace]
+[package]
+name = "q_compress"
+version = "0.7.1"
+edition = "2018"
 
 members = [
   "q_compress",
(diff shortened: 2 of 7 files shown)


mwlon:
@BartMassey Can you please update the supported data types in the README too?

There's also a section at the bottom where I mention that you can just upcast your numbers to the next supported data type (like `i16` to `i32`). Given your use case with only 3,600 numbers per file though, I can imagine storing only `i16`s in the metadata actually making a difference.

BartMassey:
Good catch. Apologies: I had missed the note in the README.

That said, I think as long as most other primitive types are supported, might as well support the whole set? For my toy application, the ergonomics are nicer with native `i16` support, and I don't see any real downside?

I added support for `u8` and `i8` in this version too, so all the current Rust integer types are now supported. Also added a note about `bool` being supported in the manpage.

mwlon:
@BartMassey please take the `i8` and `u8` types. The downsides are 1. difficulty of making converters that need to match all the cases and 2. bloated package size. Not a huge deal, but I'd rather not add such uncommon data types where the metadata overhead cost is so small.

meh:
Personally I think it makes sense to cover all the integer types, it avoids a map/collect and so on.

mwlon:
I just merged change that moved the library into a subfolder, causing the conflict here. I think it should be resolved by a simple rebase.

BartMassey:
I have reverted the `u8`/`i8` addition and got it to build and pass tests with the current structure.

Like @meh I prefer the ergonomics of `u8`/`i8` native, and don't think I understand the difficulties and costs of this? But it's ultimately your call.

mwlon:
@BartMassey Thanks for your contribution! I'm just being cautious about adding unnecessary features/data types/etc because it means more maintenance and more complexity integrating with other tools in the future. Another reason I forgot to mention earlier is that supporting `Compressor::<u8>` could confuse people into trying to use it for text-like/binary data.

PR #156, opened 2024-01-21
Add C-bindings

mrmbernardi:


pco_c/include/cpcodec.h
```diff
@@ -0,0 +1,24 @@
...
+  F64,
+} PcoType;
+
+typedef struct PcoVec {
+  const void *ptr;
+  unsigned int len;
+  const void *raw_box;
+} PcoVec;
+
+struct PcoVec auto_compress(const void *nums,
```
mwlon, line 17: Let's not address this now, but I'd like to point out that a more C-style approach to this API would be to pass in a `char *` that pco writes data to. We'd need some way to inform C how many bytes need to be allocated, though.

pco_c/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    level as usize,
+  ))
+}
+
+fn _auto_decompress<T: NumberLike>(compressed: *const u8, len: c_uint) -> PcoVec {
+  let slice = unsafe { std::slice::from_raw_parts(compressed, len as usize) };
+  let decompressed = pco::standalone::auto_decompress::<T>(slice);
+  match decompressed {
+    Ok(v) => PcoVec::from_vec(v),
+    Err(_) => PcoVec::empty(),
```
mwlon, line 60: I think we should surface an error in this case, instead of falsely returning an empty vector. I'm not super familiar with C, but perhaps it would make sense to add an error code to `PcoVec`? Alternatively, even panicking by calling `.unwrap()` on the Rust side would be good enough for now.

mwlon (changes requested):
Mostly looks good. I don't think we need to make this the perfect C API, but there are a couple small things to improve.

pco_c/include/cpcodec.h
mrmbernardi: LZ4 and Zstd for example provide a function where you give it the size of the input buffer and it gives you the worst case size of the output buffer.

In LZ4 the code for this is just `(isize) + ((isize)/255) + 16)`

pco_c/src/lib.rs
mrmbernardi: The typical thing to do would be to change the return values of all the FFI functions to return ints, have zero be success and other values represent failures defined in an enum. The structs would then instead be passed via pointers to Rust.

pco_c/include/cpcodec.h
mwlon: Yep. I think blosc just does `(size + 16)`. Somewhat different, but we have size hints; pro: the size is more conservative and tailored to the data; con: requires 3 steps (precompute stuff, get conservative size estimate, do compression). I should be able to come up with an a-priori later though formula.

pco_c/src/lib.rs
mwlon: That sounds good to me.

pco_c/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+#[repr(C)]
+pub struct PcoVec {
+  ptr: *const c_void,
+  len: c_uint,
+  raw_box: *const c_void,
+}
+
+#[repr(C)]
+pub enum PcoType {
+  U8,
```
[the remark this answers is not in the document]
mwlon: I'm on board with that approach. I think the duplication is warranted, since the enums represent different things.

pco_c/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    $(
+      impl Into<DynTypedVec> for Vec<$types> {
+        fn into(self) -> DynTypedVec { DynTypedVec::$names(self) }
+      }
+    )+
+
+    macro_rules! match_dtype {
+      ($matcher:expr, $fn:ident, $params:tt) => {
+        match $matcher {
+          $(CoreDataType::$names => $fn::<$types>$params,)+
```
mwlon, line 34: This is handy. I might add this to the main crate later.

Issue #191, opened 2024-05-16
Add a mode for quantized floats


bzinberg:
_Suggested issue labels: Feature request ("Enhancement" in GitHub iirc), Good first issue.  This is an onboarding issue for @bzinberg (me)._

## Motivation

One use case that Pco is well-positioned to add support for is "quantized floats" -- collections of floats whose values don't use the full precision available in the data type.  This happens e.g. when a data set enters a processing pipeline as `f16`s and somewhere along the way gets widened to `f32`s.  The unused precision bits in these `f32`s are all the same, and can be compressed.

## Implementation plan

@mwlon suggested the following sequence of steps:

1. Add a [synthetic dataset](https://github.com/mwlon/pcodec/tree/f9a696ff96e860cc209cd52d7a75dde3bf1a1304/pco_cli#setting-up-synthetic-data) that should be compressible under this mode, e.g., casting a bunch of `f16` normals to `f32`
2. Add the new mode to [`mode.rs`](https://github.com/mwlon/pcodec/blob/f9a696ff96e860cc209cd52d7a75dde3bf1a1304/pco/src/mode.rs), taking a `Bitlen` (alias for `u32`) as argument, and write decoding logic for [floats](https://github.com/mwlon/pcodec/blob/f9a696ff96e860cc209cd52d7a75dde3bf1a1304/pco/src/data_types/floats.rs#L211)
3. Add encoding logic. This part is more involved, but can probably reuse some logic from `float_mult`.

## Testing plan

TBD, will follow up.

mwlon:
Yep, this should substantially improve compression ratio for f16s represented as f32s in most cases. Orthogonally, we're likely to add f16s support at the API level, so you won't have to explicitly cast to f32. No one has started on this 2nd task yet - would you like to? It shouldn't be too hard.

Skielex:
Sounds promising. I'd like to help if I can, but like I said I might be a bit out of my depth here. I haven't really used Rust beyond Hello World and compression is not my field of expertise - my background is software engineering and computer vision, mainly using C-languages and Python.

That said I've been looking for an excuse to get my hands rusty, so if you don't mind, I can give it a shot. However, I'll need guidance on where to start and how you want to deal with `f16` in general, since the type isn't natively supported as far as I can tell. Also, I'm not sure exactly the work would integrate with the ongoing draft PR. I don't want to waste your time, so I'll leave it to you as to whether you think it's worth it to involve a newbie (me). 

[...]

PR #193, opened 2024-05-22
Add `FloatQuant` mode

bzinberg:
## Summary

Implements https://github.com/mwlon/pcodec/issues/191 by adding a FloatQuant / "quantized float" mode.

Adds a `f64_quantized_normal` dataset generator to `generate_randoms.py` on which we expect FloatQuant to outperform FloatMult.

Logic for automatically deciding whether to use FloatQuant over FloatMult, as well as how to automatically select the parameter `k`, are deferred to https://github.com/mwlon/pcodec/issues/194.

### Unit
Added a `split_latents` -> `join_latents` round-trip test for the new mode.

### Round trip
Not sure whether `bench` already tests this, but verified that compress -> decompress round trip leaves the data unchanged:
```
$ cargo run compress \
    --input-format csv \
    --float-mult disabled \
    --float-quant 32 \
    --input data/txt/f64_quantized_normal.txt \
    qn_float_quant.pco

$ cargo run decompress \
    --output binary \
    qn_float_quant.pco \
    > qn_float_quant_decompressed.bin

$ diff qn_float_quant_decompressed.bin \
    data/binary/f64_quantized_normal.bin
# nothing
```

### Bench
FloatQuant indeed outperforms FloatMult on the `f64_quantized_normal` dataset:
```
$ cargo run --release -- \
    bench \
    --datasets quantized_normal \
    --codecs pco,pco:float_mult=disabled:float_quant=32

╭──────────────────────┬──────────────────────────────────────────────────┬─────────────┬───────────────┬─────────────────╮
│ dataset              │ codec                                            │ compress_dt │ decompress_dt │ compressed_size │
├──────────────────────┼──────────────────────────────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ f64_quantized_normal │ pco                                              │ 51.456746ms │    8.721847ms │         3708019 │
│ f64_quantized_normal │ pco:float_mult=Disabled:float_quant=Provided(32) │ 40.630794ms │    4.655311ms │         3446617 │
│ <sum>                │ pco                                              │ 51.456746ms │    8.721847ms │         3708019 │
│ <sum>                │ pco:float_mult=Disabled:float_quant=Provided(32) │ 40.630794ms │    4.655311ms │         3446617 │
│ <sum>                │ <sum>                                            │  92.08754ms │   13.377158ms │         7154636 │
╰──────────────────────┴──────────────────────────────────────────────────┴─────────────┴───────────────┴─────────────────╯
```


bzinberg:
@mwlon, sharing this draft because I bet you'll have feedback even at this incomplete stage.

Would appreciate both high-level feedback as to "is this basically on the right track," as well as Rust tips.  Rust has been pretty interesting and not too hard to guess, but I haven't tried to compile it yet (lol) and there are probably some places that could be better expressed using language features I don't know about.

bzinberg:
Next step is to fix build errors until it compiles.

pco/src/chunk_config.rs
```diff
@@ -46,6 +46,20 @@ pub enum FloatMultSpec {
   // TODO support a LossyEnabled mode that always drops the ULPs latent var
 }
 
+/// Configures whether float-decomposition detection is enabled.
+///
+/// Examples where this helps:
+/// * float-valued data stored in a type that is unnecessarily wide (e.g. stored
+///   as `f64`s where only the first 32 bits are used)
+pub enum FloatDecompSpec {
```
mwlon, line 54: Let's call this `QuantizeSpec` or `FloatQuantizeSpec`?

pco/src/data_types/floats.rs
```diff
@@ -2,29 +2,47 @@ use std::mem;
...
 fn choose_mode_and_split_latents<F: FloatLike>(
   nums: &[F],
   chunk_config: &ChunkConfig,
 ) -> (Mode<F::L>, Vec<Vec<F::L>>) {
-  match chunk_config.float_mult_spec {
-    FloatMultSpec::Enabled => {
-      if let Some(fm_config) = float_mult_utils::choose_config(nums) {
-        let mode = Mode::float_mult(fm_config.base);
-        let latents = float_mult_utils::split_latents(nums, fm_config.base, fm_config.inv_base);
+  if chunk_config.float_decomp_spec != FloatDecompSpec::Disabled {
```
mwlon, line 11: Let's not touch this part yet. For now it will be impossible to choose Quantize mode, but that's ok.

When we do get to this, I think the logic will look more like
```
match (fm_config, quantize_config) {
  // one provided => return that
  // both provided => error
  // either enabled => do some common stuff, then maybe choose one or return classic
  // disabled => classic
}
```

pco/src/data_types/floats.rs
```diff
@@ -177,6 +197,9 @@ macro_rules! impl_float_number {
           Mode::FloatMult(base_latent) => {
             Self::from_latent_ordered(base_latent).is_finite_and_normal()
           }
+          Mode::FloatDecomp(k_latent) => {
```
mwlon, line 200: `k` isn't a latent, right? It's just a `Bitlen`? And since it's unsigned, it can't be < 0

pco/src/mode.rs
```diff
@@ -46,15 +46,32 @@ pub enum Mode<L: Latent> {
...
 
 impl<L: Latent> Mode<L> {
   pub(crate) fn n_latent_vars(&self) -> usize {
     use Mode::*;
 
     match self {
-      Classic => 1,
-      FloatMult(_) | IntMult(_) => 2,
+      Classic => 1,  // delta_order
+      FloatMult(_) | IntMult(_) => 2,  // delta_order, mode.base
```
mwlon, line 73: You can lump these cases together with another `|`

pco_cli/generate_randoms.py
```diff
@@ -144,6 +144,13 @@ def milli_micro_timestamps():
 np.random.seed(0)
 write_f64(np.random.randint(0, 2 ** 30, size=n), 'integers')
 
+# `float32`s compressed as `float64`s
+# TODO: Migrate this file to use the numpy `Generator` interface
+np.random.seed(0)
+lossless_normal = np.random.normal(size=n)
+lossy_normal = lossless_normal.astype(np.float32).astype(np.float64)
```
mwlon, line 151: quantized_normal?

mwlon:
Is there a reason you're not compiling it? I'd think it would help to have a faster feedback loop.

I think there are at least 3 more things to change (and just so you don't worry, I wouldn't expect anyone to anticipate these):

* in `pco/src/chunk_metadata.rs` we encode/decode the mode and its associated data.
* in `pco/src/constants.rs` we'll want to bump the CURRENT_FORMAT_VERSION since we'll be extending the format. This indicates that old decoders won't be able to read files written with new encoder versions.
* in `docs/format.md` we'll want to add a bit of information about this mode and the new format version

pco/src/float_decomp_utils.rs
```diff
@@ -0,0 +1,58 @@
+use std::cmp;
+
+use crate::constants::Bitlen;
+use crate::data_types::FloatLike;
+use crate::sampling;
+
+#[inline(never)]
+pub(crate) fn join_latents<F: FloatLike>(k: Bitlen, primary: &mut [F::L], secondary: &[F::L]) {
```
mwlon, line 8: (I haven't taken a close look yet, will do so tomorrow)

pco/src/mode.rs
bzinberg: Does that feel more readable to you?  I separate them out so that I could add a comment on each line saying what the parameters are.  I'll defer to you if you prefer the one-line, no-comment version though.

pco/src/chunk_config.rs
bzinberg: Will do.  Saving the mass-rename until the other stuff is resolved because mass-renames in git are a pain.

pco_cli/generate_randoms.py
bzinberg: Ditto https://github.com/mwlon/pcodec/pull/193#discussion_r1609187065

bzinberg (commented):
It compiles now.

> in `pco/src/chunk_metadata.rs` we encode/decode the mode and its associated data.

Drafted.  Haven't looked carefully to make sure I touched all the necessary places though.

> in `pco/src/constants.rs` we'll want to bump the CURRENT_FORMAT_VERSION since we'll be extending the format. This indicates that old decoders won't be able to read files written with new encoder versions.

Done.

> in `docs/format.md` we'll want to add a bit of information about this mode and the new format version

Drafted.

pco/src/float_decomp_utils.rs
```diff
@@ -0,0 +1,58 @@
...
+  for (y_and_dst, &m) in primary.iter_mut().zip(secondary.iter()) {
+    assert!(m >> k == 0, "Invalid input to FloatDecomp: m must be a k-bit integer")
```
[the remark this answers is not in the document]
bzinberg: Done.

pco_cli/generate_randoms.py
bzinberg: Done.

pco/src/data_types/floats.rs
bzinberg: I'll see what kind of lightweight feature flag makes sense when I go to add unit tests.

pco/src/mode.rs
mwlon: I didn't realize that's what you're doing - if you want to label each one, that's fine, but they'd be something like e.g. "multiplier, ULPs adjustment" or "mult, adj" for the *Mult modes.

Like for int mult mode, the compression algorithm effectively says that there are some distributions `L_0`, `L_1`  (the latent (random) variables) such that each number is `x = l_0 * base + l_1`, `l_0 ~ L_0`, `l_1 ~ L_1`. The instances `l_0, l_1` are the latents.

pco/src/data_types/floats.rs
bzinberg: Filed https://github.com/mwlon/pcodec/issues/194 to track implementing the full version of the choose logic.

pco/src/chunk_config.rs
bzinberg: Done.

pco/src/float_decomp_utils.rs
```diff
@@ -0,0 +1,58 @@
...
+    *y_and_dst = F::from_bits((*y_and_dst << k) + m).to_latent_ordered()
```
mwlon: So I thought about it and believe we want to do this a bit differently. There's quite a lot we want out of these formulas:

* The number should be monotonically increasing with the primary latent. This ensures that delta encoding the primary stream of latents makes sense and increases the smoothness of the latent distribution.
* It should be possible to write the formula in a performant way (already great!).
* Ideally the formula is easily explainable, since this will be a lasting part of the format.

Mainly the issue with the current formula is that it has a big jump and changes monotonicity direction around 0.0. I'm thinking we should do something more like this:
```rust
let quantized = F::from_latent_ordered(*y_and_dst << k);
F::from_latent_bits(quantized.to_latent_bits() + m)
```

This is order-preserving in the first latent, still pretty performant (not as much as your formula, but I think that tradeoff is well worth it for the compression ratio), and only a little bit hard to explain: "The first latent is the number of quantums away from -inf, and the second is the number of ULPs to increase magnitude by".

Do you agree?

mwlon (commented):
I've taken a look at the whole thing now. Excited! I suggest you use some `cargo` commands (and maybe get an IDE? I'm using RustRover and people like VSCode) so that you're not doing this on hard mode

bzinberg:
Thanks for the comments!  I'll take a look soon / probably tomorrow.

pco/src/mode.rs
bzinberg: > I didn't realize that's what you're doing - if you want to label each one, that's fine, but they'd be something like e.g. "multiplier, ULPs adjustment" or "mult, adj" for the *Mult modes.
> 
> Like for int mult mode, the compression algorithm effectively says that there are some distributions `L_0`, `L_1` (the latent (random) variables) such that each number is `x = l_0 * base + l_1`, `l_0 ~ L_0`, `l_1 ~ L_1`. The instances `l_0, l_1` are latents drawn from the latent (random) variables.


Ah, I didn't realize that's what `latent_vars` meant in this case, thanks!

I do think "latents" currently means several different things, and even if I'm wrong about that, it might lead to some improvements to the docs :) so I'll say a bit:

In my possibly-mistaken understanding of the codebase, "latent" is used to mean:
* A thing of type `L`, regardless of its meaning (e.g.: `to_latent_bits()`)
* The uint that corresponds to an input datum under the order-preserving injection `to_latent_ordered()`
* An element of the tuple returned by `split_latents` (an injective map that carries a single datum (e.g. an `F`) to possibly-multiple `L`s)

From talking with you, I understand there is also another notion of "latent," though I haven't seen it in the codebase:
* If we treat compression as a source coding problem, where the data are modelled as i.i.d., then pco is optimal or near-optimal in the case where the true data distribution is a certain latent variable model where the things called `l0` and `l1` in the readme are the latent variables.

I currently think of that last one as being your personal framing of pco, and it's a reasonable one.  But it seems not "official" in the sense that I don't see it reflected in the code or laid out directly in the docs, and I wouldn't expect most users to prefer that framing or arrive at it on their own.  (Maybe I'm wrong, I don't know much about the culture of compression libraries.) Also, that latent variable model is unusual in that the latents are a deterministic function of the observed value.  That's part of why I don't naturally think of pco in terms of a latent variable model, I insetad think of it in terms of carefully designed data transformations that separate out compressible components.

WDYT?

pco/src/float_decomp_utils.rs
bzinberg: Ah, I didn't think of that.  Yeah, it makes sense that we want the `split_latents` mapping, which in general already assumes that the second component is compressible, to try its best to make the first component compressible as well.

I agree there is value to having the formula be explainable in terms of abstract numbers (as opposed to being entirely based on peculiarities of the floating point format).  At this point I'm not able to assess how that compares to the value of keeping the first component of `split_latents` compressible, but I definitely see both.

I'm not sure I understand the comment about monotonicity and smoothness.  I would think that monotonicity alone doesn't imply that the pushforward of a smooth distribution by `split_latents` is also smooth, right?

bzinberg (commented):
Replied - thanks for your engagement on this stuff!

bzinberg:
> I've taken a look at the whole thing now. Excited! I suggest you use some `cargo` commands (and maybe get an IDE? I'm using RustRover and people like VSCode) so that you're not doing this on hard mode

I've been using `cargo build` and `cargo run`, and I ran `cargo test` when I first cloned the repo to make sure `main` could run on my machine.  Based on [this](https://kapeli.com/cheat_sheets/Cargo.docset/Contents/Resources/Documents/index) cheat sheet, I don't see any other obvious commands to use right now?  Definitely interested in tips tho, please hmu if applicable.

For editor, my first attempt will probably be to look for lightish-weight vim integrations with cargo, similar to what vim-fugitive does with git.  For example, it would be nice to have a vim command that runs `cargo build` and populates the quickfix window with a list of build errors that open the relevant file + line when an error is selected.

bzinberg:
Btw, I avoid VSCode specifically because I am sympathetic to the objections raised in "[Visual Studio Code is designed to fracture](https://ghuntley.com/fracture/)."

pco/src/float_decomp_utils.rs
bzinberg: Oh this is great, thank you.  I'll think about it in more detail this weekend and I bet I'll learn from it.

pco/src/mode.rs
bzinberg: Hm yeah, I probably need to understand the delta stuff and the tANS stuff more before I can make more informed comments.  One thing that crossed my mind and is maybe independent of that, wdyt having the crate-global name of the type be something like `UInt`, and then each component of the code (e.g., each `Mode`) can have a local typedef?  E.g., the existing modes could have a member `type Latent = UInt`?  This is maybe not so much a suggestion for The One Name to use, but a suggestion that maybe allowing local aliases per mode could help clarify things.

pco/src/float_decomp_utils.rs
bzinberg: Ok, I think I understand your previous comment, except a couple of details (mainly, intuitions about how benchmarks will perform) which I don't need to fully understand now.  I agree with your suggestion and will make the change, thanks!

bzinberg (commented):
Ok, next I'm gonna try adding unit tests, using the existing tests as a template.  Needless do say, I've never done this before in Rust :)

Btw,

> For editor, my first attempt will probably be to look for lightish-weight vim integrations with cargo, similar to what vim-fugitive does with git.

I installed [rust.vim](https://github.com/rust-lang/rust.vim) and it's basically this.  Now I've got a language-aware file outline in Tagbar and `RustFmt`/`RustFmtRange` available via editor command.  The `cargo` integration doesn't seem to give a clickable list of errors unfortunately, oh well.

bzinberg:
Haven't had time to work on this in the past few days, but I've got some time this evening.

Here's my estimate of what's left:
* Add unit tests
* Check with @mwlon whether I should run a non-unit-test benchmark and add a chat of the results to the PR description and/or markdown docs
* Ask @mwlon whether anything else is needed before merge

For unit tests, the obvious ones that come to mind are:
* `split_latents` → `join_latents` round trip
* FloatQuant performs better than Classic on a dataset of quantized floats

A third unit test could be sanity-check of the "decide whether to use FloatQuant" logic,
```rust
float_quant_utils::choose_config<F: FloatLike>(nums: &[F]) -> Option<FloatQuantConfig>
```
which is sort of crude and currently unused, since in this initial implementation we don't allow `Enabled`, only `Provided(k)` (see https://github.com/mwlon/pcodec/pull/193#discussion_r1609850809 and https://github.com/mwlon/pcodec/issues/194).  So it might make the most sense to just remove that code rather than unit test it for now.

bzinberg:
@mwlon, the Rust-noob errors I'm hitting are making this inefficient - I think it would be much faster (and more fun) if we pair-program on this in real time.  I'll reach out off-band to see if we can set something up.

bzinberg:
Smoke-test benchmark worked:
```
$ cargo run compress \
    --input-format csv \
    --float-mult enabled \
    --float-quant disabled \
    --input data/txt/f64_quantized_normal.txt \
    qn_float_mult.pco

$ cargo run compress \
    --input-format csv \
    --float-mult disabled \
    --float-quant 32 \
    --input data/txt/f64_quantized_normal.txt \
    qn_float_quant.pco

$ ls -sh1 qn*.pco
3.6M qn_float_mult.pco
3.3M qn_float_quant.pco
```

mwlon:
Sweet! Instead of compress, can you run `cargo run --release -- bench --datasets quantized --codecs pco,pco:float_quant=32`? AFK, apologies if typos. You might need to add a few lines into pco_cli/src/bench/codecs/pco.rs to parse float_quant

(Note to self: compress and bench subcommands should probably receive configurations in the same way in the future)

bzinberg:
> Sweet! Instead of compress, can you run `cargo run --release -- bench --datasets quantized --codecs pco,pco:float_quant=32`? AFK, apologies if typos. You might need to add a few lines into pco_cli/src/bench/codecs/pco.rs to parse float_quant
> 
> (Note to self: compress and bench subcommands should probably receive configurations in the same way in the future)

```
$ cargo run --release -- \
    bench \
    --datasets quantized_normal \
    --codecs pco,pco:float_mult=disabled:float_quant=32

╭──────────────────────┬──────────────────────────────────────────────────┬─────────────┬───────────────┬─────────────────╮
│ dataset              │ codec                                            │ compress_dt │ decompress_dt │ compressed_size │
├──────────────────────┼──────────────────────────────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ f64_quantized_normal │ pco                                              │ 51.456746ms │    8.721847ms │         3708019 │
│ f64_quantized_normal │ pco:float_mult=Disabled:float_quant=Provided(32) │ 40.630794ms │    4.655311ms │         3446617 │
│ <sum>                │ pco                                              │ 51.456746ms │    8.721847ms │         3708019 │
│ <sum>                │ pco:float_mult=Disabled:float_quant=Provided(32) │ 40.630794ms │    4.655311ms │         3446617 │
│ <sum>                │ <sum>                                            │  92.08754ms │   13.377158ms │         7154636 │
╰──────────────────────┴──────────────────────────────────────────────────┴─────────────┴───────────────┴─────────────────╯
```

bzinberg:
Just in case `bench` doesn't make sure the decompressed output is the same as the origin input, I confirmed:
```
$ cargo run compress \
    --input-format csv \
    --float-mult disabled \
    --float-quant 32 \
    --input data/txt/f64_quantized_normal.txt \
    qn_float_quant.pco

$ cargo run decompress \
    --output binary \
    qn_float_quant.pco \
    > qn_float_quant_decompressed.bin

$ diff qn_float_quant_decompressed.bin \
    data/binary/f64_quantized_normal.bin
# nothing
```

pco/src/float_quant_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+  k: Bitlen,
+  primary: &mut [F::L],
+  secondary: &[F::L],
+) -> () {
+  for (y_and_dst, &m) in primary.iter_mut().zip(secondary.iter()) {
+    debug_assert!(
+      (m >> k).is_zero(),
+      "Invalid input to FloatQuant: m must be a k-bit integer"
+    );
+    *y_and_dst = (*y_and_dst << k) + m;
```
mwlon, line 18: We wanted to change this to something that's monotonic in y, right?

pco/src/data_types/mod.rs
```diff
@@ -108,7 +109,12 @@ pub trait Latent:
   /// impossible.
   fn from_u64(x: u64) -> Self;
 
+  fn is_zero(self) -> bool {
+    self == Self::ZERO
+  }
+
   fn leading_zeros(self) -> Bitlen;
+  fn trailing_zeros(self) -> Bitlen;
```
mwlon, line 117: We've added two trailing zeros functions (here and in Latent), but I think we need just this one?

mwlon (changes requested):

pco/src/mode.rs
```diff
@@ -78,4 +106,8 @@ impl<L: Latent> Mode<L> {
   pub(crate) fn float_mult<F: FloatLike<L = L>>(base: F) -> Self {
     Self::FloatMult(base.to_latent_ordered())
   }
+
+  pub(crate) fn float_quant<F: FloatLike<L = L>>(k: Bitlen) -> Self {
```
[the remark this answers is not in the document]
bzinberg: Yep, I blindly copied from FloatMult on first pass and then forgot to remove.  Done.

pco/src/float_quant_utils.rs
```diff
@@ -0,0 +1,75 @@
...
+  fn test_join_split_round_trip() {
+    let nums = vec![1.234, -9999.999, f64::NAN, -f64::INFINITY];
+    let uints = nums
+      .iter()
+      .map(|num| num.to_latent_ordered())
+      .collect::<Vec<_>>();
+
+    let k: Bitlen = 5;
+    let mut s = split_latents(&nums, k);
+    let (l, r) = s.split_at_mut(1);
```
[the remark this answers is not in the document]
bzinberg: I tried things like that for a while and this is my understanding:
* Even when `s` is mut, `s[0]` is not mut.  You have to do `s.get_mut(0).unwrap()`.
* The borrow checker rejects `let ys = s.get_mut(0).unwrap(); let ms = s.get(1).unwrap();`, as both accesses borrow all of `s`.  This is why `split_at_mut` exists.

The slice pattern match looks great.  I tried it a few ways but haven't succeeded yet.  E.g.:
```rust
    if let [mut ys, ms] = &split_latents(&nums, k)[..] {
      join_latents::<f64>(k, &mut ys, &ms);
      assert_eq!(uints, *ys);
    } else {
      panic!("Bug: `split_latents` returned data in an unexpected format");
    }
```
gives the compile error
```
error[E0508]: cannot move out of type `[Vec<u64>]`, a non-copy slice
  --> pco/src/float_quant_utils.rs:70:27
   |
70 |     if let [mut ys, ms] = &split_latents(&nums, k)[..] {
   |             ------        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot move out of here
   |             |
   |             data moved here
   |             move occurs because `ys` has type `Vec<u64>`, which does not implement the `Copy` trait
```
Would be great to pair-program on this if we end up having the call.

pco/src/chunk_meta.rs
```diff
@@ -209,6 +209,7 @@ impl<L: Latent> ChunkMeta<L> {
     let extra_bits_for_mode = match self.mode {
       Mode::Classic => 0,
       Mode::IntMult(_) => L::BITS,
+      Mode::FloatQuant(_) => L::BITS,
```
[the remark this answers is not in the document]
bzinberg: Done.

pco/src/data_types/mod.rs
```diff
@@ -108,7 +109,12 @@ pub trait Latent:
   /// impossible.
   fn from_u64(x: u64) -> Self;
 
+  fn is_zero(self) -> bool {
```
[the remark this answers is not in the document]
bzinberg: Sure, I'm good with that.  Probably I got used to having this function from Julia.  Done.

(And re your second comment, I think this only existed for latents not floats?)

pco/src/float_quant_utils.rs
bzinberg: AIUI, it is.  Note that `y` is an `L`, as this function is only supposed to be called on the output of `split_latents`.  And in `split_latents`, there is a call to `to_latent_ordered` that makes the first output monotonic in the float representation of the input `F`.

pco/src/mode.rs
bzinberg: Oh yeah this was never for consideration in this PR.  I was suggesting in case it gave you ideas; if it doesn't sound right to you then I definitely defer to that because I don't have enough codebase familiarity to be confident that a refactor suggestion is good.

bzinberg (commented):
Thanks for the review!  Addressed most of your comments, just a couple where either I ask for more info or haven't completed yet

pco/src/chunk_config.rs
```diff
@@ -46,6 +46,21 @@ pub enum FloatMultSpec {
...
 
+/// Configures whether quantized-float detection is enabled.
+///
+/// Examples where this helps:
+/// * float-valued data stored in a type that is unnecessarily wide (e.g. stored
+///   as `f64`s where only the first 32 bits are used)
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub enum FloatQuantSpec {
+    Disabled,
+    #[default]
```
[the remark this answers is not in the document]
bzinberg: Good idea, done.

At a higher level, I was looking for a systematic way to understand this library's approach to backward compatibility.  You were able to reason about it in https://github.com/mwlon/pcodec/pull/193#discussion_r1613943528 and I remember thinking something like "should we document the backward compatibility discipline?"

pco/src/float_quant_utils.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[inline(never)]
+pub(crate) fn join_latents<F: FloatLike>(k: Bitlen, primary: &mut [F::L], secondary: &[F::L]) -> () {
+  for (y_and_dst, &m) in primary.iter_mut().zip(secondary.iter()) {
+    debug_assert!((m >> k).is_zero(), "Invalid input to FloatQuant: m must be a k-bit integer");
+    *y_and_dst = F::from_latent_bits((*y_and_dst << k) + m).to_latent_ordered();
+  }
+}
+
+pub(crate) fn split_latents<F: FloatLike>(page_nums: &[F], k: Bitlen) -> Vec<Vec<F::L>> {
```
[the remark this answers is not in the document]
bzinberg: Drafted an addition to `src/tests/recovery.rs` to include the FloatQuant case.

pco/src/float_quant_utils.rs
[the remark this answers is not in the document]
bzinberg: Got it on the call :-)

bzinberg (commented):
@mwlon, the changes from our call today are pushed.  I did a couple of small refactors, mainly of the unit test to make it cleaner to specify/modify/debug the test cases.

Ball is in your court to consider possible changes to the definition of `m` in FloatQuant for better binning behavior.  Happy to be a sounding board, of course.

pco/src/float_quant_utils.rs
mwlon: You're right. We should still solve the problem that `m` differs for perfectly quantized positive vs negative floats, though. Here are two proposals (I recommend we go with the first one):

1. We tweak flip the meaning of `m` when the float is negative. I think this would look like `let sign_cutoff = L::MID >> k; let ulps_away = if y >= cutoff {m} else {(1 << k) - 1 - m}; *dst = (y << k) + ulps_away`. This should have all the properties we want.
2. (the meaning of `y` and `m` is actually identical to 1.) We do away with the mandatory final step of `from_latent_ordered` and do `let quantums = if y >= L::MID {y - L::MID} else {L::MAX - y} ; let num = from_bits((mid_centered_quantums << k) + m)`. I think this is aesthetically better with slightly better performance, but less consistent with how I wrote the other modes, and we'd have to change some of the other code to avoid calling `from_latent_ordered`.

bzinberg: Yeah, that sounds good.  Just implemented # 1 and pushed.

Fwiw, I'm generally of the view that any trade-offs made in the name of performance should be accompanied by some kind of repeatable profiling that shows the performance gain is real -- caching and register allocation (not to mention compiler optimizations) are too complicated to be guessable in most cases.  My impression is that you do a fair bit of this sort of thing already, so I'm probably not saying anything new there :slightly_smiling_face:

As for the aesthetics stuff, one strategy I sometimes use is to have two implementations of a tricky snippet of code: in the docs, an elegantly explainable, possibly-slower reference implementation; and in the code, a performant, possibly not-easy-to-explain implementation, accompanied by a unit test confirming that the two implementations are functionally identical.

bzinberg:
Man, it is not easy to see which threads were updated!  I wonder why the review UI is this hard to use.  (it's probably hard to do well)

bzinberg:
Added a unit test confirming that `secondary == 0` in FloatQuant when the input float (whether positive or negative) is exactly quantized.

mwlon:
I think this is basically good to go! Can you fix the failing unit test? And maybe add in `-1.0 - f32::EPSILON` and `1.0 + f32::EPSILON` to it

bzinberg:
@mwlon,
> I think this is basically good to go! Can you fix the failing unit test? And maybe add in `-1.0 - f32::EPSILON` and `1.0 + f32::EPSILON` to it

Done, by replacing the intentionally-wrong `expected` with the actual.  Since the point of this unit test is to make sure the `expected`s make sense to us, could you sanity check them before merging?  They look good to me.

mwlon (approved):

mwlon:
Thanks for your hard work!

PR #197, opened 2024-05-29
16-bit data types

mwlon:


pco/src/data_types/floats.rs
```diff
@@ -147,7 +150,121 @@ macro_rules! impl_float_number {
         l as Self
       }
     }
+  };
+}
+
+impl FloatLike for f16 {
```
Skielex, line 156: I wonder if separate `impl FloatLike for f16`, and by extension splitting of the `impl_float_like` macro, could've been avoided. I ran into a bunch of issues related to the use of `as` to convert to and from `f16` and well as simple assignments, e.g., `const ZERO: Self = 0.0;` and operations, e.g., `const MAX_FOR_SAMPLING: Self = Self::MAX * 0.5;`.

pco/src/data_types/floats.rs
```diff
@@ -147,7 +150,121 @@ macro_rules! impl_float_number {
...
+  const BITS: Bitlen = 16;
+  const PRECISION_BITS: Bitlen = Self::MANTISSA_DIGITS as Bitlen - 1;
+  const ZERO: Self = f16::ZERO;
+  const MAX_FOR_SAMPLING: Self = f16::from_bits(30719); // Half of MAX size.
+
+  #[inline]
+  fn abs(self) -> Self {
```
Skielex, line 163: Perhaps we should instead use:
``` rust
    fn abs(self) -> Self {
        Self::from_bits(self.to_bits() & 0x7FFF)
    }
```
Seems like it should be faster. I'm a little bit confused as to why `half` has two different implementations of `abs`:
https://docs.rs/half/latest/src/half/num_traits.rs.html#221
https://docs.rs/half/latest/src/half/num_traits.rs.html#386

Perhaps it would be even better to simply use the `abs` implementation from `half`. However, I couldn't figure out how to do that.

Skielex:
@mwlon, you'll probably want to check the `header_byte` for `u16`, `i16`, and `f16`.

pco_cli/src/dtypes.rs
```diff
@@ -143,6 +144,49 @@ parquetable!(f64, parquet::data_type::DoubleType, "DOUBLE");
 parquetable!(i32, parquet::data_type::Int32Type, "INT32");
 parquetable!(i64, parquet::data_type::Int64Type, "INT64");
 
+impl Parquetable for f16 {
```
Skielex, line 147: Not sure what we wonna do here.

pco/src/data_types/floats.rs
Skielex: Perhaps a better solutions would be to rely on `num_traits` which provides implementations for `abs`, `inv` (named `recip`), `round`, `exp2`, `from_f64`, `to_f64`, `max`, and `min`. What do you think?

pco/src/data_types/floats.rs
```diff
@@ -147,7 +150,121 @@ macro_rules! impl_float_number {
...
+  }
+
+  #[inline]
+  fn to_f64(self) -> f64 {
+    self.to_f64()
+  }
+
+  #[inline]
+  fn is_finite_and_normal(&self) -> bool {
+    self.is_finite() && !(matches!(self.classify(), FpCategory::Subnormal))
```
mwlon, line 193: I think f16 provides `is_normal()`?

pco/src/data_types/floats.rs
mwlon: I agree this should be faster. I think the code you pasted should work? What error are you getting?

pco/src/data_types/floats.rs
```diff
@@ -147,7 +150,121 @@ macro_rules! impl_float_number {
...
+  const ZERO: Self = f16::ZERO;
+  const MAX_FOR_SAMPLING: Self = f16::from_bits(30719); // Half of MAX size.
+
+  #[inline]
+  fn abs(self) -> Self {
+    Self::from_f32(self.to_f32().abs())
+  }
+
+  fn inv(self) -> Self {
+    Self::from_f32(1.0 / self.to_f32())
```
mwlon, line 168: I think these `from_f32(... self.to_f32())` ones can be rewritten to use native f16 ops when available. Like I think this one could be `Self::ONE / self`.

pco_c/Cargo.toml
```diff
@@ -8,5 +8,6 @@ name = "cpcodec"
 crate-type = ["dylib"]
 
 [dependencies]
+half = { version = "2.4.1", features = ["std"] }
```
mwlon, line 11: I don't think we should need this explicitly?

pco_cli/Cargo.toml
```diff
@@ -22,8 +22,14 @@ better_io = { version = "0.1.0", path = "../better_io" }
 blosc-src = { version = "0.3", features = ["zstd"], optional = true }
 bytes = "1.1.0"
 clap = { version = "4.5.3", features = ["derive"] }
+half = { version = "2.4.1", features = ["std"] }
```
mwlon, line 25: Yeah we shouldn't need to add these transitive dependencies

pco_cli/src/parse.rs
```diff
@@ -33,8 +33,10 @@ pub fn arrow_dtype(s: &str) -> anyhow::Result<DataType> {
   let name_pairs = [
     ("f32", DataType::Float32),
     ("f64", DataType::Float64),
+    ("i16", DataType::Int16),
```
mwlon, line 36: We'll want to add f16 here too

pco_python/src/standalone.rs
```diff
@@ -66,7 +66,7 @@ pub fn register(_py: Python, m: &PyModule) -> PyResult<()> {
   ///
   /// :param nums: numpy array to compress. This may have any shape.
   /// However, it must be contiguous, and only the following data types are
-  /// supported: float32, float64, int32, int64, uint32, uint64.
+  /// supported: float16, float32, float64, int16, int32, int64, uint16, uint32, uint64.
```
mwlon, line 69: Nice. We'll also want to update the main README

pco_cli/src/dtypes.rs
mwlon: Let's just leave it as a TODO for now. This is only used for benchmarks, and we can add support in later.

Can you run `cargo clippy`? I think you'll get some warnings that variables are unused here and should be prefixed with `_`.

mwlon:
We'll also want to add float 16 to `pco_cli/src/arrow_handlers.rs`

Skielex:
> We'll also want to add float 16 to `pco_cli/src/arrow_handlers.rs`

Done

pco/src/data_types/floats.rs
mwlon: In the channel I said:

> Num traits might make sense, but would be a big change affecting more than just this. We'd have to experiment and see how it changes binary size. Let's not consider it for now

I still think this is right, and we should explore it in a separate branch later. This PR is already quite broad, and we shouldn't change too many things at once.

pco/src/data_types/floats.rs
Skielex: Perhaps a better solutions would be to rely on `num_traits` which provides implementations for `abs`, `inv` (named `recip`), `round`, `exp2`, `from_f64`, `to_f64`, `max`, and `min`. What do you think?

mwlon (commented):
Looks mostly good, just needs some adjustments!

pco/src/data_types/floats.rs
Skielex: Not getting any errors with the code i pasted. Pushed the change now. The issue was when i tried calling `self.abs()`. I suppose it's because we don't have the `num_traits`. For now we'll just use our own `Self::from_bits(self.to_bits() & 0x7FFF)` implementation.

pco/src/data_types/floats.rs
Skielex: Sorry, realized I'd put the comment the wrong place, so I moved it, only to see that you'd already replied. We'll keep `num_traits` out of this for now.

pco/src/tests/recovery.rs
```diff
@@ -72,13 +73,45 @@ fn assert_recovers<T: NumberLike>(
...
-  assert_recovers(&[f64::MIN, f64::MAX], 0, "float extremes 0")?;
-  assert_recovers(&[1.2_f32], 0, "float 0")?;
-  assert_recovers(&[1.2_f32], 1, "float 1")?;
-  assert_recovers(&[1.2_f32], 2, "float 2")?;
-  assert_recovers(&Vec::<u32>::new(), 6, "empty 6")?;
-  assert_recovers(&Vec::<u32>::new(), 0, "empty 0")
+  assert_recovers(
+    &[u64::MIN, u64::MAX],
+    0,
+    "u64 extremes 0 - 0",
```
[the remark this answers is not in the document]
Skielex: Not sure. Removing extra `0 `.

pco_cli/Cargo.toml
Skielex: It won't compile without it because `pco_cli` relies on macros from `dynamics`:
```
error[E0433]: failed to resolve: use of undeclared crate or module `half`
  --> pco_cli\src\core_handlers.rs:25:3
   |
25 |   with_core_dtypes!(match_dtype)
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `half`
   |
   = note: this error originates in the macro `with_core_dtypes` (in Nightly builds, run with -Z macro-backtrace for more info)
```

pco/src/tests/recovery.rs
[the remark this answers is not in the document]
Skielex: I think sluijs added the `-`. Let me know if you want it removed.

pco_c/Cargo.toml
Skielex: The `half` dep is needed for compilation. See other reply.

pco/src/data_types/floats.rs
Skielex: Probably, though I'm not sure what exactly will happen at the lower level since I'm not sure the CPU has a native `f16` divide op. I just copied the [implementation from `half`](https://docs.rs/half/latest/src/half/num_traits.rs.html#269), but I'll change this to `Self::ONE / self`.

Got any suggestions from the others? Again I just copied from `half`:
https://docs.rs/half/latest/src/half/num_traits.rs.html#206
https://docs.rs/half/latest/src/half/num_traits.rs.html#436

Skielex:
Thanks for the feedback. Think I covered it all now. Let me know if I forgot something.

pco_cli/Cargo.toml
mwlon: Ah right, now it makes sense that we actually need it here

mwlon:
@Skielex congrats on your first PR!

PR #203, opened 2024-06-13
Added missing options to `PyChunkConfig`, fixes #202

### LINKED ISSUE #202: Add missing options to `PyChunkConfig`
Several options and properties are missing from `PyChunkConfig`. This includes:
- The `float_quant_spec` property.
- The `int_mult_spec` `Provided` option.
- The `float_mult_spec` `Provided` option.

I think updating `PyChunkConfig` to include all `ChunkConfig` options should be relatively simple. Should we do it?

Skielex:
# Summary
This PR adds missing options to the `PyChunkConfig` and aligns it with `ChunkConfig`. This is done by adding three new wrapper classes. 

`PyFloatMultSpec` and `PyFloatQuantSpec` are used for the `int_mult_spec` and `float_mult_spec` properties instead of `String`. This allows `base` values to passed in using the `provided` function. `Enabled` and `Disabled` are now set using the `enabled` and `disabled` functions, rather than passing in strings.

The `float_quant_spec` property of type `PyFloatQuantSpec` has been added to `PyChunkConfig`. This works similar to the other two wrappers.

This fixes #202.

# Tasks done
- [x] Add `PyIntMultSpec` class
- [x] Add `PyFloatMultSpec` class
- [x] Add `PyFloatQuantSpec` class
- [x] Update `PyChunkConfig` to use the three classes above
- [x] Update docstrings
- [x] Add Python tests for new options

### DIFF
--- a/pco_python/src/lib.rs
+++ b/pco_python/src/lib.rs
@@ -167,11 +167,11 @@ impl PyChunkConfig {
   ///
   /// :param compression_level: a compression level from 0-12, where 12 takes
   /// the longest and compresses the most.
-  /// 
+  ///
   /// :param delta_encoding_order: either a delta encoding level from 0-7 or
   /// None. If set to None, pcodec will try to infer the optimal delta encoding
   /// order.
-  /// 
+  ///
   /// :param int_mult_spec: a IntMultSpec that configures whether integer
   /// multiplier detection is enabled.
   ///
@@ -184,7 +184,7 @@ impl PyChunkConfig {
   /// substantially reduced. This configuration may hurt
   /// compression speed slightly even when it isn't helpful.
   /// However, the compression ratio improvements tend to be large.
-  /// 
+  ///
   /// :param float_mult_spec: a FloatMultSpec that configures whether float
   /// multiplier detection is enabled.
   ///
@@ -197,14 +197,14 @@ impl PyChunkConfig {
   /// substantially reduced. In rare cases, this configuration
   /// may reduce compression speed somewhat even when it isn't helpful.
   /// However, the compression ratio improvements tend to be large.
-  /// 
+  ///
   /// :param float_quant_spec: a FloatQuantSpec that configures whether
   /// quantized-float detection is enabled.
   ///
   /// Examples where this helps:
   /// * float-valued data stored in a type that is unnecessarily wide (e.g.
   /// stored as `f64`s where only a `f32` worth of precision is used)
-  /// 
+  ///
   /// :param paging_spec: a PagingSpec describing how many numbers should
   /// go into each page.
   ///
--- a/pco_python/src/lib.rs
+++ b/pco_python/src/lib.rs
@@ -54,24 +54,24 @@ pub fn pco_err_to_py(pco: PcoError) -> PyErr {
 #[derive(Clone, Default)]
 pub struct PyIntMultSpec(IntMultSpec);
 
-/// Specifies how pcodec should handle modulo compression for integer types.
+/// Specifies if pcodec should use an integer multiplier to compress data.
 #[pymethods]
 impl PyIntMultSpec {
-  /// :returns: a IntMultSpec disabling modulo compression.
+  /// :returns: a IntMultSpec disabling integer multiplier detection.
   #[staticmethod]
   fn disabled() -> Self {
     Self(IntMultSpec::Disabled)
   }
 
-  /// :returns: a IntMultSpec enabling modulo compression. This will
-  /// automatically try to detect whether to use the mode and which `base` to
-  /// use.
+  /// :returns: a IntMultSpec enabling integer multiplier detection. This
+  /// will automatically try to detect whether to use the mode and which `base`
+  /// to use.
   #[staticmethod]
   fn enabled() -> Self {
     Self(IntMultSpec::Enabled)
   }
 
-  /// :returns: a IntMultSpec with a specific `base` for modulo compression.
+  /// :returns: a IntMultSpec with a specific `base` to use for compression.
   #[staticmethod]
   fn provided(base: u64) -> Self {
     Self(IntMultSpec::Provided(base))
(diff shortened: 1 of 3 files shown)


Skielex:
@mwlon , I tried following the structure used for `PyPagingSpec`. I hope it makes sense.

I'm getting this warning related to `clone()`. Not sure what the proper solution is here.
```
warning: using `clone` on type `IntMultSpec` which implements the `Copy` trait
   --> pco_python\src\lib.rs:210:27
    |
210 | ...ec(py_config.int_mult_spec.0.clone...
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `py_config.int_mult_spec.0`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
    = note: `#[warn(clippy::clone_on_copy)]` on by default
   --> pco_python\src\lib.rs:212:30    |
212 | ...ec(py_config.float_quant_spec.0.clone...
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try removing the `clone` call: `py_config.float_quant_spec.0`      
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
```

I also think it's better if someone better qualified than me writes the docstrings.

mwlon:
@Skielex This looks like the right approach. The error messages are correct:

> try removing the `clone` call

https://doc.rust-lang.org/std/clone/trait.Clone.html

Thanks for the PR!

Skielex:
Oh silly me. Thought I was getting other warnings when I removed `clone`, but clearly those "warning: assigning the result of `Clone::clone()` may be inefficient" were completely unrelated.

Skielex:
@mwlon, how do you want the Python files formatted? Perhaps we could specify the tool and config in `pyproject.toml`? I used default `black` here for the test files, but clearly you've been using something else.

mwlon:
I haven't been using any Python formatter so far. I somewhat prefer 2 spaces per indent, but I'm willing to cave on that if other people have strong opinions. I'd suggest
* option A: set up a config file in this PR with 2 spaces per indent so the diff in this PR isn't so large
* option B: make a separate PR changing only the formatting (to be the default)

In either case, I'm hoping we can generate the other .py files in this repo too (pco_cli/generate_randoms.py).

Skielex:
Sounds good. I've reverted the style changes as much as I could and created a new issue #205.

Skielex:
> I haven't been using any Python formatter so far. I somewhat prefer 2 spaces per indent, but I'm willing to cave on that if other people have strong opinions. I'd suggest
> 
> * option A: set up a config file in this PR with 2 spaces per indent so the diff in this PR isn't so large
> * option B: make a separate PR changing only the formatting (to be the default)
> 
> In either case, I'm hoping we can format the other .py files in this repo too (pco_cli/generate_randoms.py).

@mwlon , let's go with option B.

pco_python/test/test_standalone.py
```diff
@@ -92,13 +113,52 @@ def test_compression_options():
...
+              delta_encoding_order=1,
+              int_mult_spec=IntMultSpec.disabled(),
+              float_mult_spec=FloatMultSpec.disabled(),
+              float_quant_spec=FloatQuantSpec.disabled(),
+              paging_spec=PagingSpec.equal_pages_up_to(77),
+          ),
+      )) > default_size)
+
+
+@pytest.mark.parametrize("int_mult_spec", all_int_mult_specs)
```
mwlon, line 130: Hmm, I don't think this really tests that the right configuration is being used. I think instead we should just consider a distribution where the mode is definitely beneficial and confirm we get better sizes using `enabled()` or `provided(correct_answer)`.

To get such a distribution for mult modes, you could set `scale=777` (probably anything >500 would work) in the `np.random.normal` call.

Really the best thing would be based on chunk metadata instead of compressed size, but the Python API doesn't support inspecting chunk metadata yet.

pco_python/src/lib.rs
```diff
@@ -92,12 +166,21 @@ impl PyChunkConfig {
   /// :param delta_encoding_order: either a delta encoding level from 0-7 or
   /// None. If set to None, pcodec will try to infer the optimal delta encoding
   /// order.
-  /// :param int_mult_spec: either 'enabled' or 'disabled'. If enabled, pcodec
-  /// will consider using int mult mode, which can substantially improve
-  /// compression ratio but decrease speed in some cases for integer types.
-  /// :param float_mult_spec: either 'enabled' or 'disabled'. If enabled, pcodec
-  /// will consider using float mult mode, which can substantially improve
-  /// compression ratio but decrease speed in some cases for float types.
+  /// :param int_mult_spec: a IntMultSpec disabling, enabling, or providing a
```
mwlon, line 169: I don't think we should list the available options here - better for people to look at the `*Spec` docs for that.

pco_python/src/lib.rs
```diff
@@ -92,12 +166,21 @@ impl PyChunkConfig {
...
+  /// base for modulo compression. If enabled, pcodec will consider using int
+  /// mult mode, which can substantially improve compression ratio but decrease
+  /// speed in some cases for integer types. If a base is provided, pcodec will
+  /// use that base for modulo compression. Enabled by default.
```
mwlon, line 173: I think saying "modulo compression" invents a new term that we'd also need to explain. And in general I think we should put this sort of explanation in the docs for the `.provided()` function.

pco_python/src/lib.rs
```diff
@@ -92,12 +166,21 @@ impl PyChunkConfig {
...
-  /// :param int_mult_spec: either 'enabled' or 'disabled'. If enabled, pcodec
-  /// will consider using int mult mode, which can substantially improve
-  /// compression ratio but decrease speed in some cases for integer types.
-  /// :param float_mult_spec: either 'enabled' or 'disabled'. If enabled, pcodec
-  /// will consider using float mult mode, which can substantially improve
-  /// compression ratio but decrease speed in some cases for float types.
+  /// :param int_mult_spec: a IntMultSpec disabling, enabling, or providing a
+  /// base for modulo compression. If enabled, pcodec will consider using int
+  /// mult mode, which can substantially improve compression ratio but decrease
+  /// speed in some cases for integer types. If a base is provided, pcodec will
```
mwlon, line 172: Let's give some hint as to what cases these are. E.g. in the Rust docs we give a couple examples: https://docs.rs/pco/latest/pco/enum.IntMultSpec.html

pco_python/src/lib.rs
```diff
@@ -50,6 +50,79 @@ pub fn pco_err_to_py(pco: PcoError) -> PyErr {
...
+/// Specifies how pcodec should handle modulo compression for integer types.
+#[pymethods]
+impl PyIntMultSpec {
+  /// :returns: a IntMultSpec disabling modulo compression.
+  #[staticmethod]
+  fn disabled() -> Self {
+    Self(IntMultSpec::Disabled)
+  }
+
+  /// :returns: a IntMultSpec enabling modulo compression.
```
mwlon, line 66: Let's say that this will automatically try to detect whether to use the mode and which `base` to use? (we should start doing this in the Rust docs too)

Skielex: Changing it to
```
  /// :returns: a IntMultSpec enabling modulo compression. This will
  /// automatically try to detect whether to use the mode and which `base` to
  /// use.

pco_python/src/lib.rs
Skielex: Will copy the explanations from the Rust docs.

pco_python/src/lib.rs
Skielex: Changed

pco_python/src/lib.rs
Skielex: Changed

pco_python/test/test_standalone.py
Skielex: The tests were primarily added to check that all the configs work in the sense that they don't crash or return empty. I've changed the tests to make the intent more clear.

If you'd like the tests to check that the different options work as expected for specific distributions, perhaps you can fix that? I'm not sure how to do it properly.

pco_python/src/lib.rs
Skielex: Fixing up a few more places.

pco_python/test/test_standalone.py
mwlon: Ok, this seems fine for now

mwlon:
Thanks @Skielex !

PR #204, opened 2024-06-14
f16 dataset, 16-bit wavs, fixed f16 naming in bench subcommand

mwlon:
* added f16_normal to generate_randoms.py
* interpreted most wav data as 16-bit instead of 32-bit now
* fixed issue where `bench` printed out the fully qualified type name for f16 (half): `│ half::binary16::f16_normal │ pco   │   19.8805ms │    1.790916ms │         1691993 │`


mwlon:
FYI @Skielex - some minor quality of life improvements

PR #206, opened 2024-06-15
Corrected or simplified docs, especially regarding modes

mwlon:


mwlon:
@bzinberg I simplified one one of the docs of mine I noticed in your PR, and also corrected the formula which changed part-way through your PR. Might be worthwhile for you to take a quick look

bzinberg:
Thanks for the ping, I took a look.  Nice improvements.  I think we should
keep an eye on whether the intentionally-slightly-wrongness of the formulas
is sufficiently called out, given that users may read just small fragments
of the full docs.

Issue #194, opened 2024-05-23
Revise "choose mode" logic to accommodate another float mode


bzinberg:
Tracking issue for the point raised by @mwlon in https://github.com/mwlon/pcodec/pull/193#discussion_r1609193001:

> Let's not touch this part yet. For now it will be impossible to choose Quantize mode, but that's ok.
> 
> When we do get to this, I think the logic will look more like
> ```
> match (fm_config, quantize_config) {
>   // one provided => return that
>   // both provided => error
>   // either enabled => do some common stuff, then maybe choose one or return classic
>   // disabled => classic
> }
> ```

[Currently](https://github.com/mwlon/pcodec/blob/ad6560b1b564b0619476aeee874fc067bddac868/pco/src/data_types/floats.rs#L7-L29), the logic is "Check whether using `FloatMult` looks promising.  If so, use it; otherwise, use `Classic`."  As of https://github.com/mwlon/pcodec/pull/193, there is a second float-specific mode, `FloatQuant`, and there will also be logic to guess whether using `FloatQuant` looks promising.  So the "either enabled" case in the above quote needs some way to weigh these two opinions against each other (or do its own separate logic).

bzinberg:
Removed this draft of the `choose_mode` logic from https://github.com/mwlon/pcodec/pull/193 because we decided to save it for when this issue is addressed in full:
https://github.com/mwlon/pcodec/blob/15a06c4be18d9278800c7e4c98842893de7b862f/pco/src/float_quant_utils.rs#L48-L80

bzinberg:
Now that https://github.com/mwlon/pcodec/pull/193/ is merged, I think it makes sense to clarify requirements and scope on this.  My current thoughts are:

* FloatQuant is really for the case where all or part of a dataset is exactly quantized, i.e., the `secondary` component is exactly zero for a significant portion of the dataset.
* That's why the logic I linked above in https://github.com/mwlon/pcodec/issues/194#issuecomment-2142684001 searches for the largest `k` for which 90% of the dataset is exactly quantized.  The 90% was arbitrary, can be changed.
* When both FloatMult and FloatQuant are enabled, we need a way to decide which one to use.  We of course don't want tons of "regressions" -- common use cases in which automatically choosing FloatQuant over FloatMult makes performance worse -- though shooting for absolutely no regressions on any dataset is also not feasible/useful.
[...]

PR #212, opened 2024-06-19
Add `FloatQuantSpec::Enabled`

bzinberg:
Fixes https://github.com/mwlon/pcodec/issues/194.


bzinberg:
@mwlon, can you sanity check that this has the right shape?  If good then I'll add unit tests.

mwlon:
@bzinberg At a high level this looks right, but I think we should change 2 things:

* We should probably structure the match statements like
  * either_provided => obvious thing
  * any_combination_of_enabled_disabled => have each enabled mode make an optional "bid", put those in a Vec, take the max bid by bits saved or else return classic
* Mode selection takes a nontrivial amount of time, so we should try to reuse as much compute as we can between the modes. I think we should at least reuse the `sample` we choose, and I'll leave it up to you whether there's anything worth reusing with the trailing_zeros part.

Side note: I'm kinda regretting making these separate enums. Since we'll be releasing 0.3.0 next anyway, maybe I/we should make a single enum `ModeSpec` with variants like `ProvidedClassic, ProvidedIntMult(base), ..., AutoDetect`. Probably a separate PR though.

bzinberg:
Ok sounds good.

For the reusing compute point, I was guessing mode selection was light enough to not justify the amount of re-plumbing necessary to reuse `sample`.  Thanks for chiming in, I'll change it.

> Side note: I'm kinda regretting making these separate enums. Since we'll be releasing 0.3.0 next anyway, maybe I/we should make a single enum `ModeSpec` with variants like `ProvidedClassic, ProvidedIntMult(base), ..., AutoDetect`. Probably a separate PR though.

There are some decisions to make about this interface too, of course.  Would `AutoDetect` let the user specify which modes to consider?  Then mode specification becomes a two-step process: first compute a bid for each mode under consideration, then call a function that takes in a `Vec<Bid>`.  If a bid is always an estimate of bits saved per element, then that last function has an obvious implementation; not sure if that will stay true.

I'll ping back when I have unit tests to look over (likely tomorrow eve).

Skielex:
> Side note: I'm kinda regretting making these separate enums. Since we'll be releasing 0.3.0 next anyway, maybe I/we should make a single enum `ModeSpec` with variants like `ProvidedClassic, ProvidedIntMult(base), ..., AutoDetect`. Probably a separate PR though.

If the three modes are exclusive (as I understand they are), I think this change would make a lot of sense. This way we can help the user avoid most invalid or non-sensical combinations of specs. 
However, to completely prevent the user from being able to use `ProvidedFloatMult` when compressing integers, I think we'll have to use two different types of `ChunkConfig`, e.g., `IntChunkConfig` and `FloatChunkConfig`. The type of config should then match the type of input data.

This should make it impossible for the user to accidentally provide a `ProvidedFloatMult` when compressing integers, and vice versa.

Could be something along the lines of:
`IntChunkConfig` extends `ChunkConfig` and has a `IntModeSpec` property.
`FloatChunkConfig` extends `ChunkConfig` and has a `FloatModeSpec` property.
`IntModeSpec` has options `ProvidedClassic`, `ProvidedMult(base)`, `AutoDetect`.
`FloatModeSpec` has options `ProvidedClassic`, `ProvidedMult(base)`, `ProvidedQuant(base)`, `AutoDetectMultOnly`, `AutoDetectQuantOnly`, `AutoDetect`.

Not sure if this a better solution, or if it's even practical to implement given the heavy use of templating.

mwlon:
@bzinberg good question - I think for the initial release we'll just make `AutoDetect` monolithic, but we should probably leave open the later possibility of choosing a subset of modes. So really it would be something like

```rust
#[non_exhaustive]
pub struct AutoDetectConfig {
  // empty for now, could add power user features later
}

pub enum ModeSpec {
  ProvidedClassic,
  ...,
  AutoDetect(AutoDetectConfig),
}
```

That said, I think structuring this automatic detection as something like `vec![bid_a, bid_b].iter().max_by_key(|bid| bid.bits_saved)` in any case.

@Skielex Yep, the modes are exclusive. I see what you're getting at, but I think the complexity of making `ChunkConfig` generic would require a lot more code and outweight the benefits. For similar reasons, I've made e.g. `FloatMultSpec::Provided(f64)` instead of `...Provided(T)`. On incorrect use, users will get `InvalidArgument` errors, which are obviously not as good as compile-time errors, but these will still fail fast and be easy to test for.

bzinberg:
Filed https://github.com/mwlon/pcodec/issues/215 for the ModeSpec stuff, and since we're trying to cut the 0.3.0 release soon, we can handle that afterwards.

bzinberg:
Speaking of which, I was unexpectedly occupied yesterday, returning to this today.

bzinberg:
> * We should probably structure the match statements like
> 
>   * either_provided => obvious thing
>   * any_combination_of_enabled_disabled => have each enabled mode make an optional "bid", put those in a Vec, take the max bid by bits saved or else return classic

We can simplify the type signatures here: instead of a vector of `Option<Bid>`s, let's just do a vector of `Bid`s and have there be a type of bid that says "don't choose me."
 
> * Mode selection takes a nontrivial amount of time, so we should try to reuse as much compute as we can between the modes. I think we should at least reuse the `sample` we choose, and I'll leave it up to you whether there's anything worth reusing with the trailing_zeros part.

I would expect mode selection to take much less time than any of the work that touches the entire dataset, so I'm not sure what use case is motivating what you say here.

If we do really want to reuse as much work as possible in all of the combinatorially-many choices of modes to consider, then we're really talking about a build system.

mwlon:
I think at least the sample is going to be common to any (non-classic) mode we consider. Maybe I'm being too perfectionist, but with float mult at present, ~7% of compression time is spent choosing the mode (whether or not we end up using float mult). We take a small sample, but we do relatively complicated things to it. I fully expect to increase that 7%, but let's take easy wins where we can.

bzinberg:
Ah yeah sounds good, we can definitely reuse the sample. I don't expect
much improvement from that alone, since FloatQuant does only very simple
stuff to the sample, but the general point about wanting to have the code
structured to be able to reuse work (with reusing the sample as a starting
point) makes sense, will do.

pco/src/data_types/mod.rs
```diff
@@ -16,6 +16,8 @@ mod floats;
 mod signeds;
 mod unsigneds;
 
+pub(crate) type ModeAndLatents<L> = (Mode<L>, Vec<Vec<L>>);
```
bzinberg, line 19: perhaps document what the `Vec<Vec<L>>` means?  Something like
```suggestion
/// Container type for (i) a mode, and (ii) the result of transforming a dataset using that mode
/// (see `split_latents`).  The number of elements in the outer `Vec` equals the number of latents
/// that each (float) element of the dataset is split into, and the length of each inner `Vec` is
/// the number of elements in the dataset.
pub(crate) type ModeAndLatents<L> = (Mode<L>, Vec<Vec<L>>);
```

pco/src/float_mult_utils.rs
```diff
@@ -651,25 +628,37 @@ mod test {
...
         base: 1.0 / 7.0,
         inv_base: 7.0,
       })
     );
-    assert_eq!(choose_config(&ones), None);
+    assert_eq!(
+      choose_config(&ones),
+      Some(FloatMultConfig {
+        base: 1.0,
+        inv_base: 1.0,
```
bzinberg, line 644: Are there any unit tests that would fail if `inv_base` was not correctly plumbed through the library code (i.e. got clobbered by `base.inv()` somewhere)?

pco/src/float_quant_utils.rs
```diff
@@ -54,48 +56,46 @@ pub(crate) fn split_latents<F: FloatLike>(page_nums: &[F], k: Bitlen) -> Vec<Vec
...
+  // will usually be used on datasets that are exactly quantized.
+  let bits_saved_per_num = (k as f64) * freq;
   if (k as f64) > QUANT_REQUIRED_BITS_SAVED_PER_NUM {
-    Bid::Candidate {
+    Some(Bid {
       mode: Mode::FloatQuant(k),
-      bits_saved_per_num: k as f64,
-    }
+      bits_saved_per_num,
+      split_fn: Box::new(move |nums| split_latents(nums, k)),
```
bzinberg, line 69: Note to self: ah, gotta look up `move` -- probably related to some combination of `std::move` and rvalue references in C++ but I bet there are differences

pco/src/float_quant_utils.rs
```diff
@@ -54,48 +56,46 @@ pub(crate) fn split_latents<F: FloatLike>(page_nums: &[F], k: Bitlen) -> Vec<Vec
...
+    rev_csum += occurrences;
+    if rev_csum >= thresh {
+      return (
+        k as Bitlen,
+        rev_csum as f64 / sample.len() as f64,
+      );
+    }
+  }
+
+  unreachable!("nums should be nonempty")
```
bzinberg, line 98: I'm surprised so little revision was needed, since I hadn't run the code in a unit test yet lol.

For my own edification, do you recall what went wrong in the first draft?

pco/src/float_quant_utils.rs
```diff
@@ -104,6 +104,28 @@ mod test {
...
+  fn test_estimate_best_k() {
+    // all but the last of these have 21 out of 23 mantissa bits zeroed
+    let mut sample = vec![1.0_f32, 1.25, -1.5, 1.75, -0.875, 0.75, 0.625].repeat(3);
+    sample.push(f32::from_bits(1.0_f32.to_bits() + 1));
+    let (k, freq) = estimate_best_k_and_freq(&sample);
+    assert_eq!(k, 21);
+    assert_eq!(
+      freq,
+      (sample.len() - 1) as f64 / (sample.len() as f64)
+    );
```
bzinberg, line 117: I'd suggest explicitly showing the dependence on `REQUIRED_QUANTIZED_PROPORTION` so that readers don't miss it:
```suggestion
    );
    assert!(freq >= REQUIRED_QUANTIZED_PROPORTION);
```

pco/src/float_quant_utils.rs
```diff
@@ -104,6 +104,28 @@ mod test {
...
+  }
+
+  #[test]
+  fn test_estimate_best_k_full_precision() {
+    // all but the last of these have 21 out of 23 mantissa bits zeroed
```
bzinberg, line 122:
```suggestion
    // all elements have all 52 mantissa bits zeroed
```

pco/src/mode.rs
```diff
@@ -103,11 +103,10 @@ impl<L: Latent> Mode<L> {
...
-  #[default]
-  Forfeit,
-  /// Bid to use this mode (TODO: elaborate)
-  Candidate { mode: Mode<L>, bits_saved_per_num: f64 },
+pub(crate) struct Bid<T: NumberLike> {
+  pub mode: Mode<T::L>,
+  pub bits_saved_per_num: f64,
+  // we include a split_fn since modes like FloatMult can benefit from extra
+  // information (inv_base) not captured entirely in the mode
+  pub split_fn: Box<dyn FnOnce(&[T]) -> Vec<Vec<T::L>>>,
```
bzinberg, line 111: This makes sense for how it's being used right now.

I'm wondering if there is an implicit API change/refinement that should be clarified somewhere.  In particular, as a user of the library I would wonder: if `split_latents` receives side info besides the `Mode`, then what do I need to know about that side info in order to use the library correctly?

I think the answer to the above is: users do not need to know anything about the side info; it is an internal detail of the compression algorithm that allows it to auto-detect parameters more effectively and, in particular, the decompression step does not need to know anything about it.

Come to think of it, maybe the place to doc that is right here.
```suggestion
  // we include a split_fn since modes like FloatMult can benefit from extra
  // information (inv_base) not captured entirely in the mode.  This extra
  // information is an implementation detail of the compressor, not part of the
  // format itself, and is not / does not need to be known to the decompressor.
  pub split_fn: Box<dyn FnOnce(&[T]) -> Vec<Vec<T::L>>>,
```

pco/src/data_types/floats.rs
```diff
@@ -6,144 +6,98 @@ use crate::constants::Bitlen;
...
-        let latents = float_mult_utils::split_latents(nums, base, base.inv());
-        Ok((mode, latents))
-      }
-      Bid::Candidate { .. } => Err(PcoError::internal(
-        "Bug: expected a FloatMult bid, got a different mode",
-      )),
-      Bid::Forfeit => Ok((Mode::Classic, split_latents_classic(nums))),
-    },
-    (_, FloatQuantSpec::Enabled) => match sampling::choose_sample(
+    _ => Ok(detect_mode_and_split_latents(
```
bzinberg, line 89: Looks like the casework on `Spec`s is split between this function and `detect_mode_and_split_latents`.  Would it make sense to put them all in one place, e.g. by handling `Provided` in the bid creation step?

bzinberg (commented):
Good stuff, thanks for picking this up!

pco/src/data_types/floats.rs
```diff
@@ -173,7 +127,7 @@ macro_rules! impl_float_like {
 
       #[inline]
       fn exp2(power: i32) -> Self {
-        Self::from_bits(((-$exp_offset + power) as $latent) << Self::PRECISION_BITS)
+        Self::from_bits((($exp_offset + power) as $latent) << Self::PRECISION_BITS)
```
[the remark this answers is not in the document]
mwlon: Yes, please disregard

pco/src/data_types/mod.rs
mwlon: Since this is pub(crate) it won't be exposed to users - they'll see the the full type instead of the alias. How about we document `choose_mode_and_split_latents` instead?

pco/src/float_quant_utils.rs
mwlon: I think the idea is that it takes ownership of any variables it needs for the closure, as opposed to capturing them by reference (in this case `k`).

pco/src/float_quant_utils.rs
mwlon: Sure!

* the hist went from `[0, PRECISION_BITS)`, but we could have trailing zeros in `[0, PRECISION_BITS]`, so we need the + 1
* the csum omitted the current element (addition should come before the thresh check)
* I think you would have wanted to take `.first()` instead of `.last()`. For me it's a little easier to reason about as a loop that terminates as soon as it hits the condition.

pco/src/data_types/floats.rs
mwlon: Would we just use a +1.0 or +Inf bid? I kinda like this approach better; seems a tad roundabout to make bids when we know which one will win.

pco/src/float_mult_utils.rs
mwlon: Hmm, probably not, but I have unit tests for `snap_to_int_reciprocal`. To test this we'd want some multiples of e.g. 1/7, some of which are extremely large such that 1.0/7.0 ~= 1ULP

pco/src/mode.rs
mwlon: Oh I absolutely do not intend to expose this to users (even the power users who might consider looking at `NumberLike`. It's only `pub(crate)`.

bzinberg: Ah yeah, I meant to say users and maintainers -- imo it's worth saying this here so that readers don't have to think hard to know it.  Is there reason not to include it?

pco/src/float_quant_utils.rs
bzinberg: Ah cool.  I would have expected each argument to have its own such qualifier (as iirc they do in C++) rather than a single one for all of them.

pco/src/float_mult_utils.rs
bzinberg: Do you think it's worth adding?

pco/src/data_types/mod.rs
bzinberg: Ah yeah, typedefs in general are hard to document for this reason, true.  Adding a sentence like the above to the doc of `choose_mode_and_split_latents` sounds good.

pco/src/data_types/floats.rs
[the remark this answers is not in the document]
bzinberg: Disregard meaning they'll be removed from the PR (possibly by hitting `main` and then merging `main` into this branch) before merge?

pco/src/float_quant_utils.rs
bzinberg: Ah that's true, I initialized `hist` too small, and this would have resulted in an out-of-bounds assignment if any of the `x`s had all their mantissa bits equal to zero.

The csum lagging by one element was intentional, and taking `last()` was also intentional -- we are iterating over all $r$ such that
$$\big| \lbrace x \ |\ \text{TZ}(x) \geq r \rbrace \big| < \texttt{thresh}$$
and we want one before the smallest such $r$.

I'm fine with the rewrite though.  You're probably right that reasoning about the `rev` of an `enumerate` is more cognitive load for most readers than reasoning about an ascending index that gets subtracted from stuff.

pco/src/data_types/floats.rs
```diff
@@ -6,52 +6,99 @@ use crate::constants::Bitlen;
...
+  if num.is_finite_and_normal() {
+    let abs = num.abs();
+    if abs <= F::MAX_FOR_SAMPLING {
+      return Some(abs);
+    }
+  }
+  None
+}
+
+fn detect_mode_and_split_latents<F: FloatLike>(
```
bzinberg, line 29: We should add a unit test like the one you outlined in https://github.com/mwlon/pcodec/issues/194#issuecomment-2153285188, where `FloatMult::Enabled` and `FloatQuant::Enabled` are competing against each other and it's a close call.

pco/src/data_types/floats.rs
bzinberg: Oh right, that's a good reason to have separate code places -- if there's no auction, then handle it right here, and if there is an auction, then create the bids.  Sounds good.

pco/src/float_quant_utils.rs
bzinberg: oh, the rewrite also uses the `rev` of an `enum`.  oh well lol, still fine

pco/src/data_types/floats.rs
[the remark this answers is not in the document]
mwlon: Done

pco/src/data_types/mod.rs
mwlon: Done

pco/src/data_types/floats.rs
```diff
@@ -424,6 +424,29 @@ impl_float_number_like!(f16, u16, 1_u16 << 15, 9);
...
 
+  #[test]
+  fn test_choose_mult_mode() {
+    let base = 1.7;
+    let nums = (0..300).map(|i| (i as f64) * base).collect::<Vec<_>>();
+    let (mode, _) = choose_mode_and_split_latents(&nums, &ChunkConfig::default()).unwrap();
+    assert_eq!(
+      mode,
+      Mode::FloatMult(base.to_latent_ordered())
+    );
```
bzinberg, line 435: It's a coincidence that we get exact equality when `base = 1.7`, right?  In general there may be rounding error?

pco/src/float_quant_utils.rs
```diff
@@ -207,4 +214,27 @@ mod test {
...
+    assert_eq!(bid.bits_saved_per_num, 17.0);
+
+    // same as above, except not all perfectly quantized
+    let mut sample = (0..100).map(|x| x as f32).collect::<Vec<_>>();
+    sample[0] += 0.1;
+    sample[37] -= 0.1;
+    let bid = compute_bid(&sample).unwrap();
+    assert!(matches!(bid.mode, Mode::FloatQuant(17)));
+    assert!(bid.bits_saved_per_num < 17.0);
+    assert!(bid.bits_saved_per_num > 15.0);
```
bzinberg, line 233: Nice

pco_cli/src/chunk_config_opt.rs
```diff
@@ -25,7 +25,7 @@ pub struct ChunkConfigOpt {
   #[arg(long, default_value = "Enabled", value_parser = parse::float_mult)]
   pub float_mult: FloatMultSpec,
   // TODO(https://github.com/mwlon/pcodec/issues/194): Implement "Enabled" mode
```
bzinberg, line 27: Time to remove this todo

pco/src/chunk_config.rs
```diff
@@ -55,6 +55,9 @@ pub enum FloatMultSpec {
 pub enum FloatQuantSpec {
   #[default]
   Disabled,
+  /// Automatically decide whether to use FloatQuant, based on an estimate of how much it would
+  /// save compared to other available modes.
+  Enabled,
   /// If you know how many bits of precision are unused, you can supply that
   /// number here.
   Provided(Bitlen),
```
bzinberg, line 63: GH won't let me comment on the line below this but there is a todo on lines 61-62 that it's time to remove

pco/src/float_quant_utils.rs
```diff
@@ -52,12 +57,82 @@ pub(crate) fn split_latents<F: FloatLike>(page_nums: &[F], k: Bitlen) -> Vec<Vec
...
+  let (k, freq) = estimate_best_k_and_freq(sample);
+  // Nothing fancy, we simply estimate that quantizing by k bits results in saving k bits per
+  // number, whenever possible. This is based on the assumption that FloatQuant
+  // will usually be used on datasets that are exactly quantized.
+  let bits_saved_per_infrequent_primary = freq * (k as f64);
+  let bits_saved_per_num = sampling::est_bits_saved_per_num(
+    sample,
+    |x| x.to_latent_bits() >> k,
+    bits_saved_per_infrequent_primary,
+  );
```
bzinberg, line 70: Not for this PR but, are there tests that make sure this estimate is close to the actual saved bits of compression?

bzinberg (commented):
Looks good to merge once the minor comments below are addressed!

pco/src/float_quant_utils.rs
mwlon: There are not. We could maybe add these for well-behaved cases? In antagonistic or weird cases our estimates might be substantially off.

pco/src/data_types/floats.rs
mwlon: The heuristics are usually pretty good, but I changed it to 1.5 for more exact arithmetic.

PR #217, opened 2024-06-28
Float quant enabled

mwlon:


bzinberg:
Looks like there was some kind of git hiccup pushing to #212?  Messaged you off band to try to figure it out.  In any case, I just added your commits there to preserve comment history.

PR #220, opened 2024-07-03
ModeSpec, replacing invididual specs for each mode

### LINKED ISSUE #215: Refactor the `FooSpec`s to be mutually exclusive
> @bzinberg good question - I think for the initial release we'll just make `AutoDetect` monolithic, but we should probably leave open the later possibility of choosing a subset of modes. So really it would be something like
> 
> ```rust
> #[non_exhaustive]
> pub struct AutoDetectConfig {
>   // empty for now, could add power user features later
> }
> 
> pub enum ModeSpec {
>   ProvidedClassic,
>   ...,
>   AutoDetect(AutoDetectConfig),
> }
> ```
> 
> That said, I think structuring this as something like `vec![bid_a, bid_b].iter().max_by_key(|bid| bid.bits_saved)` makes sense in any case.
> 
> @Skielex Yep, the modes are exclusive. I see what you're getting at, but I think the complexity of making `ChunkConfig` generic would require a lot more code and outweight the benefits. For similar reasons, I've made e.g. `FloatMultSpec::Provided(f64)` instead of `...Provided(T)`. On incorrect use, users will get `InvalidArgument` errors, which are obviously not as good as compile-time errors, but these will still fail fast and be easy to test for.
> 
_Originally posted by @mwlon in https://github.com/mwlon/pcodec/issues/212#issuecomment-2183543586_

mwlon:
closes #215

### DIFF
--- a/pco/src/data_types/floats.rs
+++ b/pco/src/data_types/floats.rs
@@ -27,39 +27,29 @@ fn filter_sample<F: FloatLike>(num: &F) -> Option<F> {
   None
 }
 
-fn auto_mode_and_split_latents<F: FloatLike>(
-  nums: &[F],
-  chunk_config: &ChunkConfig,
-) -> ModeAndLatents<F::L> {
-  // up to 3 bids: classic, float mult, float quant modes
-  let mut bids: Vec<Bid<F>> = vec![];
-  bids.push(Bid {
-    mode: Mode::Classic,
-    bits_saved_per_num: 0.0,
-    split_fn: Box::new(|nums| split_latents_classic(nums)),
-  });
-
-  if matches!(chunk_config.mode_spec, ModeSpec::Auto) {
-    if let Some(sample) = sampling::choose_sample(nums, filter_sample) {
-      bids.extend(float_mult_utils::compute_bid(&sample));
-      bids.extend(float_quant_utils::compute_bid(&sample));
-    }
-  }
-
-  let winning_bid = choose_winning_bid(bids);
-  let latents = (winning_bid.split_fn)(nums);
-  (winning_bid.mode, latents)
-}
-
 fn choose_mode_and_split_latents<F: FloatLike>(
   nums: &[F],
   chunk_config: &ChunkConfig,
 ) -> PcoResult<ModeAndLatents<F::L>> {
   match chunk_config.mode_spec {
-    ModeSpec::Auto => Ok(auto_mode_and_split_latents(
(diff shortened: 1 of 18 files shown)


mwlon:
FYI @bzinberg. Very satisfying to delete so many LOC - thanks for enduring the previous incarnation of this spec!

pco/src/tests/recovery.rs
```diff
@@ -52,21 +52,11 @@ fn assert_recovers<T: NumberLike>(
...
-    for (fm, fq) in [
-      (
-        FloatMultSpec::Enabled,
-        FloatQuantSpec::Disabled,
-      ),
-      (
-        FloatMultSpec::Disabled,
-        FloatQuantSpec::Provided(14),
-      ),
-    ] {
```
bzinberg, line 64: Why the behavior change on this unit test?

pco/src/data_types/floats.rs
```diff
@@ -435,17 +413,16 @@ mod tests {
...
-  // #[test]
-  // fn test_choose_quant_mode() {
-  //   let lowest_num_bits = 1.0_f64.to_bits();
-  //   let k = 20;
-  //   let nums = (0..1000)
-  //     .map(|i| f64::from_bits(lowest_num_bits + (i << k)))
-  //     .collect::<Vec<_>>();
-  //   let (mode, _) = choose_mode_and_split_latents(&nums, &ChunkConfig::default()).unwrap();
-  //   assert_eq!(mode, Mode::FloatQuant(k));
-  // }
```
bzinberg, line 448: How did this test get fixed?  I'm not seeing it, but if the PR contains code that fixes the failing test, can we add that information to the PR description?

bzinberg (commented):
Oh yeah this rules.  I'm confused though, see comments

pco/src/data_types/floats.rs
mwlon: I'm being confusing: this was actually fixed in https://github.com/mwlon/pcodec/pull/219, and I should have uncommented the test there, but forgot.

pco/src/tests/recovery.rs
mwlon: At least the first spec here is no longer possible, and I think `Auto + Classic` cover this pretty well. But I might have reduced some coverage with the 2nd one, so maybe I/we should add a new recovery test to hit that case. Thanks for the post-hoc review.

PR #221, opened 2024-07-03
Revise minimum versions of `half` and `bytes`

LDeakin:
The `half` minimum version is relaxed from 2.4.1 to 2.0.0 for most of the `pco` crates.

For `pco_cli`, `half` is relaxed to 2.1.0 and `bytes` increased from 1.1.0 to 1.4.0.
These minimum versions are set by the `arrow` related deps.

### DIFF
--- a/pco_cli/Cargo.toml
+++ b/pco_cli/Cargo.toml
@@ -22,7 +22,7 @@ better_io = { version = "0.1.0", path = "../better_io" }
 blosc-src = { version = "0.3", features = ["zstd"], optional = true }
 bytes = "1.1.0"
 clap = { version = "4.5.3", features = ["derive"] }
-half = { version = "2.4.1", features = ["std"] }
+half = { version = "2.1.0", features = ["std"] }
 indicatif = "0.17.8"
 parquet = { version = "49.0.0", features = [
   "arrow",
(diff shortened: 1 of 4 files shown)


pco_cli/Cargo.toml
```diff
@@ -20,9 +20,9 @@ anyhow = "1.0.71"
 arrow = { version = "49.0.0", features = ["csv"], default-features = false }
 better_io = { version = "0.1.0", path = "../better_io" }
 blosc-src = { version = "0.3", features = ["zstd"], optional = true }
-bytes = "1.1.0"
+bytes = "1.4.0"
```
mwlon, line 23: I understand why we're reducing version requirements, but is there a reason to increase this one?

LDeakin: It isn't strictly necessary, but dependency resolution will always choose version 1.4.0+ due to the other dependencies.

LDeakin: FYI, I used [cargo-minimal-versions](https://crates.io/crates/cargo-minimal-versions) (with the top-level workspace `Cargo.toml` [removed](https://github.com/taiki-e/cargo-minimal-versions/issues/1)) to check that each `pco` crate builds with [`-Z direct-minimal-versions`](https://crates.io/crates/cargo-minimal-versions).

mwlon:
@LDeakin do you need a new release with this change?

LDeakin:
Yes, but there is no rush.

mwlon:
@LDeakin this is released in 0.3.1 now.

PR #224, opened 2024-07-18
Simpler min-max functions that vectorize

Skielex:
# Background
The current `slice_min` and `slice_max` functions are not very clean, [nor do they seem to vectorize](https://rust.godbolt.org/z/537xGvj8r).

# Changes
I've replaced the for-loops with `latents.iter().fold()`, which appears to be vectorized by the compiler.

I've added a new `slice_min_max` function that gets both values in one iteration. To use this effectively I've had to make a change to `apply_incomplete`, which I admit is not the cleanest. Suggestions are welcome!

# Performance
~~Honestly, I'm not able to see any repeatable change in performance when running the benchmarks. I've done some digging and it appears the time spent in the min-max functions is so little that I cannot properly benchmark it.~~

~~@mwlon, do you know any scenarios which will cause more time to be spent in these functions that we could use to benchmark this properly? If not, then I don't think there's much point spending more time here.~~

Running 
```
cargo run --release -- bench -i data/binary/ --input-format binary --codecs pco:level=0 --iters 100
```
this PR appears to net ~5% reduction in total time, while 
```
cargo run --release -- bench -i data/binary/ --input-format binary --codecs pco:level=2 --iters 100
```
is ~2% faster. For higher compression levels the effect appears to be insignificant.

### DIFF
--- a/pco/src/histograms.rs
+++ b/pco/src/histograms.rs
@@ -44,30 +44,22 @@ pub struct HistogramBin<L: Latent> {
 }
 
 fn slice_min<L: Latent>(latents: &[L]) -> L {
-  let mut min_val = L::MAX;
-  for i in 0..latents.len() {
-    min_val = min(min_val, latents[i]);
-  }
-  min_val
+  latents
+    .iter()
+    .fold(L::MAX, |min_val, val| min(min_val, *val))
 }
 
 fn slice_max<L: Latent>(latents: &[L]) -> L {
-  let mut max_val = L::ZERO;
-  for i in 0..latents.len() {
-    max_val = max(max_val, latents[i]);
-  }
-  max_val
+  latents
+    .iter()
+    .fold(L::ZERO, |max_val, val| max(max_val, *val))
 }
 
 fn slice_min_max<L: Latent>(latents: &[L]) -> (L, L) {
-  let mut min_val = L::MAX;
-  let mut max_val = L::ZERO;
-  for i in 0..latents.len() {
-    let val = latents[i];
-    min_val = min(min_val, val);
-    max_val = max(max_val, val);
-  }
-  (min_val, max_val)
+  latents.iter().fold(
+    (L::MAX, L::ZERO),
+    |(min_val, max_val), val| (min(min_val, *val), max(max_val, *val)),
+  )
 }
 
 struct HistogramBuilder<L: Latent> {
--- a/pco/src/histograms.rs
+++ b/pco/src/histograms.rs
@@ -44,29 +44,30 @@ pub struct HistogramBin<L: Latent> {
 }
 
 fn slice_min<L: Latent>(latents: &[L]) -> L {
-  let mut min0 = L::MAX;
-  let mut min1 = L::MAX;
-  for i in (0..latents.len()).skip(1).step_by(2) {
-    min0 = min(min0, latents[i - 1]);
-    min1 = min(min1, latents[i]);
+  let mut min_val = L::MAX;
+  for i in 0..latents.len() {
+    min_val = min(min_val, latents[i]);
   }
-  if latents.len() % 2 == 1 {
-    min0 = min(min0, latents.last().cloned().unwrap());
-  }
-  min(min0, min1)
+  min_val
 }
 
 fn slice_max<L: Latent>(latents: &[L]) -> L {
-  let mut max0 = L::ZERO;
-  let mut max1 = L::ZERO;
-  for i in (0..latents.len()).skip(1).step_by(2) {
-    max0 = max(max0, latents[i - 1]);
-    max1 = max(max1, latents[i]);
+  let mut max_val = L::ZERO;
+  for i in 0..latents.len() {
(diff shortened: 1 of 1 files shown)


mwlon:
Do we still get vectorization with `.iter().min().unwrap()` instead of fold?

Skielex:
> Do we still get vectorization with `.iter().min().unwrap()` instead of fold?

No, `latents.iter().min().unwrap().clone()` doesn't seem to vectorize.

I benchmarked with `cargo run --release -- bench -i data/binary/ --input-format binary --codecs pco:level=0 --iters 100` and this PR seems to net ~5% reduction in total time.

pco/src/histograms.rs
```diff
@@ -101,18 +94,22 @@ impl<L: Latent> HistogramBuilder<L> {
...
+      bin.upper = match upper {
+        Bound::Loose(_) => slice_max(latents),
+        Bound::Tight(upper) => upper,
+      };
       bin.count += latents.len();
     } else {
-      let tight_lb = match lower {
-        Bound::Loose(_) => slice_min(latents),
-        Bound::Tight(lower) => lower,
+      let (tight_lb, tight_ub) = match lower {
```
mwlon, line 104: Can you make this a single match on (lower, upper)?

pco/src/histograms.rs
```diff
@@ -44,29 +44,22 @@ pub struct HistogramBin<L: Latent> {
...
-  let mut min1 = L::MAX;
-  for i in (0..latents.len()).skip(1).step_by(2) {
-    min0 = min(min0, latents[i - 1]);
-    min1 = min(min1, latents[i]);
-  }
-  if latents.len() % 2 == 1 {
-    min0 = min(min0, latents.last().cloned().unwrap());
-  }
-  min(min0, min1)
+  latents
```
mwlon, line 47: I find that we can do `latents.iter().cloned().min().unwrap()` and still get SIMD. I think the difference between this and `*latents.iter().min().unwrap()` is that the latter operates on references to latents instead of values.

mwlon (changes requested):

pco/src/histograms.rs
Skielex: Oh, you can match on tuples? Sweet!

Do you have a insights on how often each match arm might be hit so we can order them accordingly?

pco/src/histograms.rs
Skielex: Didn't realise a had to clone before aggregating for it to vectorize. Though now that I look at it, it's what I was doing with `fold`.

One of the reasons I was using fold is that it was functionally equivalent to the previous implementation in the way it would handle an empty array. I also kinda like that it avoids `unwrap`. `fold` also appears to produce slightly fewer instructions, though I doubt it makes any measurable difference.

If we make `slice_min` and `slice_max` panic on empty array, then I think `slice_min_max` should panic too. If we do it like this
``` rust
fn slice_min<L: Latent>(latents: &[L]) -> L {
  latents.iter().cloned().min().unwrap()
}

fn slice_max<L: Latent>(latents: &[L]) -> L {
  latents.iter().cloned().max().unwrap()
}

fn slice_min_max<L: Latent>(latents: &[L]) -> (L, L) {
  latents.iter().cloned().fold(
    (L::MAX, L::ZERO),
    |(min_val, max_val), val| (min(min_val, val), max(max_val, val)),
  )
}
```
then the three functions handle empty arrays differently, which isn't ideal imho. I don't know if there's a clever way to do `slice_min_max` that makes it function like the two others.

mwlon: I don't think you should worry about it, since these functions are private to this module and only used on nonempty slices. But if this still bothers you, you could do e.g. `.unwrap_or(L::MAX)` to make slice_min behave like `slice_min_max`.

pco/src/histograms.rs
mwlon: I don't think it matters 😅 This part isn't super hot, and branch prediction will handle it.

But to satisfy your curiosity: we're almost always doing quicksort here, typically pivoting by x from [a, b) into [a, x) and [x, b). So most often it's tight LB, loose UB.

Skielex: Ok, I'll just leave them in alphabetical order then.

pco/src/histograms.rs
Skielex: Personally, I think the current implementation where all three functions use `fold` is cleaner than the alternatives, as it sticks to one way of doing things.

If we got rid of the functions and aggregated inline I think using the built-in `min` and `max` funcitons would be cleaner. However, as long as we use the custom `slice_` functions, which I think makes the code more readable overall, I like that they're consistent, both functionally and implementation wise.

It's your call of course :-)
If you prefer something else, feel free to push the changes.

mwlon (approved):

pco/src/histograms.rs
mwlon: Ok, seems reasonable. I just made a tiny simplification.

PR #225, opened 2024-07-18
Replaced `log2()` with `log(2.0)` on Windows, fixes #223

### LINKED ISSUE #223: `f32::log2` very slow on Windows
For some reason the `f32::log2()` function, which is used a lot through `bin_cost` is very slow on Windows. I collected some data with Intel VTune running the following benchmark:
```
bench -i .\data\binary --input-format binary --codecs pco:level=8,pco:level=0 --iters 100 --dtypes f16,f32,f64
```
Here's a bottom-up view of the time spent in various functions:
image

We can see that more than 8 seconds (14% of the total time) is spent in `fdlog`. By comparison, if we change from `f32::log2()` to `f32::log(2.0)`, which calls `logf` instead of `fdlog`, only about 0.7 second (1.4% of the total time) is spent getting the base-2 log.

Here's the bottom-up view when using `f32::log(2.0)` instead:
image

The issue can be reproduced on my AMD 5900X and Intel 1370P, both running Windows 11. On Linux this doesn't not appear to be an issue.

I'm not sure what the best remedy is. We could simply replace `f32::log2()` with `f32::log(2.0)` for Windows.

Skielex:
This PR replaces `log2()` with `log(2.0)` on Windows. The motivation is that `log2()` is *very* slow on Windows making `log(2.0)` about 10x faster.

See #223 for more details.

### DIFF
--- a/pco/src/bin_optimization.rs
+++ b/pco/src/bin_optimization.rs
@@ -16,6 +16,11 @@ fn bin_cost<L: Latent>(
   total_count_log2: f32,
 ) -> f32 {
   let count = count as f32;
+  // On Windows, log2() is very slow, so we use log(2.0) instead, which is
+  // about 10x faster. On other platforms, we stick with log2(). See #223.
+  #[cfg(target_os = "windows")]
+  let ans_cost = total_count_log2 - count.log(2.0);
+  #[cfg(not(target_os = "windows"))]
   let ans_cost = total_count_log2 - count.log2();
   let offset_cost = bits::bits_to_encode_offset(upper - lower) as f32;
   bin_meta_cost + (ans_cost + offset_cost) * count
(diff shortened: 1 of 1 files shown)


pco/src/bin_optimization.rs
```diff
@@ -42,6 +47,11 @@ fn choose_optimized_partitioning<L: Latent>(
   let total_count = c;
   let lowers = bins.iter().map(|bin| bin.lower).collect::<Vec<_>>();
   let uppers = bins.iter().map(|bin| bin.upper).collect::<Vec<_>>();
+  // On Windows, log2() is very slow, so we use log(2.0) instead, which is
+  // about 10x faster. On other platforms, we stick with log2(). See #223.
+  #[cfg(target_os = "windows")]
```
mwlon, line 52: This is only run once per compression, so let's not complicate the code in this part.

mwlon (approved):

PR #231, opened 2024-07-31
Better WAV-file support, fixes #230

### LINKED ISSUE #230: Problems with multichannel WAV files
I tried compressing a stereo WAV file, which produced the error `invalid number of bytes per wav file sample`, even though the samples are 16-bit.

I think we should use `bits_per_sample` instead of `bytes_per_sample` in the `get_wav_field` function.

# Background
Header read in the `get_wav_field` function looks like this:
```
audio_format = 1
channel_count = 2
sampling_rate = 48000
bytes_per_second = 192000
bytes_per_sample = 4
bits_per_sample = 16
```
Notice `bytes_per_sample` and `bits_per_sample` values.

The data type is read from the header as:
``` rust
  let dtype = match header.bytes_per_sample {
    1 | 2 => Ok(DataType::Int16),
    3 => Ok(DataType::Int32),
    4 => Ok(DataType::Float32),
    _ => Err(anyhow!(
      "invalid number of bytes per wav file sample"
    )),
  }?;
```
While the `Header` field is named `bytes_per_sample`, it appears it should rather be named `bytes_per_block`. See header spec: https://en.wikipedia.org/wiki/WAV#WAV_file_header
```
   BytePerBloc     (2 bytes) : Number of bytes per block (NbrChannels * BitsPerSample / 8).
   BitsPerSample   (2 bytes) : Number of bits per sample
```

Changing the code to
``` rust
  let dtype = match header.bits_per_sample {
    8 | 16 => Ok(DataType::Int16),
    24 => Ok(DataType::Int32),
    32 => Ok(DataType::Float32),
    _ => Err(anyhow!(
      "invalid number of bits per wav file sample"
    )),
  }?;
```
allows `compress` to run without errors.

Skielex:
Check WAV header audio format and bit per sample to select proper data type for WAV samples. Fixes issues with multichannel WAV files #230.

### DIFF
--- a/pco_cli/src/input/audio.rs
+++ b/pco_cli/src/input/audio.rs
@@ -11,12 +11,26 @@ pub fn get_wav_field(path: &Path) -> Result<Option<Field>> {
   // this is excessively slow, but easy for now
   let mut file = File::open(path)?;
   let (header, _) = wav::read(&mut file)?;
-  let dtype = match header.bytes_per_sample {
-    1 | 2 => Ok(DataType::Int16),
-    3 => Ok(DataType::Int32),
-    4 => Ok(DataType::Float32),
+  let dtype = match header.audio_format {
+    wav::WAV_FORMAT_PCM => match header.bits_per_sample {
+      8 | 16 => Ok(DataType::Int16),
+      24 | 32 => Ok(DataType::Int32),
+      _ => Err(anyhow!(
+        "unsupported bit depth for WAV_FORMAT_PCM: {}",
+        header.bits_per_sample
+      )),
+    },
+    wav::WAV_FORMAT_IEEE_FLOAT => match header.bits_per_sample {
+      16 => Ok(DataType::Float16),
+      32 => Ok(DataType::Float32),
+      _ => Err(anyhow!(
+        "unsupported bit depth for WAV_FORMAT_IEEE_FLOAT: {}",
+        header.bits_per_sample
+      )),
+    },
     _ => Err(anyhow!(
-      "invalid number of bytes per wav file sample"
+      "unsupported audio format: {}",
+      header.audio_format
     )),
   }?;
   let name = path
--- a/pco_cli/src/input/audio.rs
+++ b/pco_cli/src/input/audio.rs
@@ -14,18 +14,11 @@ pub fn get_wav_field(path: &Path) -> Result<Option<Field>> {
   let dtype = match (header.audio_format, header.bits_per_sample) {
     (wav::WAV_FORMAT_PCM, 8 | 16) => Ok(DataType::Int16),
     (wav::WAV_FORMAT_PCM, 24 | 32) => Ok(DataType::Int32),
-    (wav::WAV_FORMAT_PCM, _) => Err(anyhow!(
-      "unsupported bit depth for WAV_FORMAT_PCM: {}",
-      header.bits_per_sample
-    )),
     (wav::WAV_FORMAT_IEEE_FLOAT, 32) => Ok(DataType::Float32),
-    (wav::WAV_FORMAT_IEEE_FLOAT, _) => Err(anyhow!(
-      "unsupported bit depth for WAV_FORMAT_IEEE_FLOAT: {}",
-      header.bits_per_sample
-    )),
     _ => Err(anyhow!(
-      "unsupported audio format: {}",
-      header.audio_format
+      "audio format {} with {} bits per sample not supported",
+      header.audio_format,
+      header.bits_per_sample
     )),
   }?;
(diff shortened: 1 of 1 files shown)


pco_cli/src/input/audio.rs
```diff
@@ -11,12 +11,26 @@ pub fn get_wav_field(path: &Path) -> Result<Option<Field>> {
...
+    wav::WAV_FORMAT_PCM => match header.bits_per_sample {
+      8 | 16 => Ok(DataType::Int16),
+      24 | 32 => Ok(DataType::Int32),
+      _ => Err(anyhow!(
+        "unsupported bit depth for WAV_FORMAT_PCM: {}",
+        header.bits_per_sample
+      )),
+    },
+    wav::WAV_FORMAT_IEEE_FLOAT => match header.bits_per_sample {
+      16 => Ok(DataType::Float16),
```
mwlon, line 24: Is this a real possibility? The `wav` crate doesn't allow decoding 16 bit floats (see `get_array` below), so I think we should only handle the int16, int32, and float32 cases.

pco_cli/src/input/audio.rs
```diff
@@ -11,12 +11,26 @@ pub fn get_wav_field(path: &Path) -> Result<Option<Field>> {
   // this is excessively slow, but easy for now
   let mut file = File::open(path)?;
   let (header, _) = wav::read(&mut file)?;
-  let dtype = match header.bytes_per_sample {
-    1 | 2 => Ok(DataType::Int16),
-    3 => Ok(DataType::Int32),
-    4 => Ok(DataType::Float32),
+  let dtype = match header.audio_format {
```
mwlon, line 14: Can we match both audio_format and bits_per_sample at once instead of nesting?

pco_cli/src/input/audio.rs
Skielex: Ok, I'll remove `16 => Ok(DataType::Float16),`.

pco_cli/src/input/audio.rs
Skielex: Definitely! Thought it was necessary to get accurate error messages at first, but this is ofc much nicer.

pco_cli/src/input/audio.rs
```diff
@@ -11,12 +11,21 @@ pub fn get_wav_field(path: &Path) -> Result<Option<Field>> {
...
+      header.bits_per_sample
+    )),
+    (wav::WAV_FORMAT_IEEE_FLOAT, 32) => Ok(DataType::Float32),
+    (wav::WAV_FORMAT_IEEE_FLOAT, _) => Err(anyhow!(
+      "unsupported bit depth for WAV_FORMAT_IEEE_FLOAT: {}",
+      header.bits_per_sample
+    )),
     _ => Err(anyhow!(
-      "invalid number of bytes per wav file sample"
+      "unsupported audio format: {}",
```
mwlon, line 27: Can we just have one failure case instead of 3, formatting in the format (with debug if necessary, `{:?}`).

Skielex: Sure, like this?
``` rust
  let dtype = match (header.audio_format, header.bits_per_sample) {
    (wav::WAV_FORMAT_PCM, 8 | 16) => Ok(DataType::Int16),
    (wav::WAV_FORMAT_PCM, 24 | 32) => Ok(DataType::Int32),
    (wav::WAV_FORMAT_IEEE_FLOAT, 32) => Ok(DataType::Float32),
    _ => Err(anyhow!(
      "unsupported audio format: {:?}",
      header
    )),
  }?;
```
Just thought it'd be more user friendly to provide a more accurate error message, but if you prefer to keep the code simpler that's fine with me :)

Skielex: Ok, technically they're both included above since the entire header is printed.

But we can print only `audio_format` and `bits_per_sample` if you prefer. Is this what you're looking for?
``` rust
  let dtype = match (header.audio_format, header.bits_per_sample) {
    (wav::WAV_FORMAT_PCM, 8 | 16) => Ok(DataType::Int16),
    (wav::WAV_FORMAT_PCM, 24 | 32) => Ok(DataType::Int32),
    (wav::WAV_FORMAT_IEEE_FLOAT, 32) => Ok(DataType::Float32),
    _ => Err(anyhow!(
      "audio format {} with {} bits per sample not supported",
      header.audio_format,
      header.bits_per_sample
    )),
  }?;
```
Otherwise, I think it's easier if you just fix it ;-)

mwlon (approved):

mwlon:
Thanks for your work!

PR #235, opened 2024-10-05
Changed C-binding data length to `size_t`

Skielex:
This removes arbitrary data size restriction from C-interface.

### DIFF
--- a/pco_c/include/cpcodec_generated.h
+++ b/pco_c/include/cpcodec_generated.h
@@ -7,18 +7,18 @@ typedef enum PcoError {
 
 typedef struct PcoFfiVec {
   const void *ptr;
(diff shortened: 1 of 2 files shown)


Skielex:
Should we change the other interface types to fixed width types (e.g., `int32_t`) while we're at it? I don't see why not.

mwlon:
@Skielex that sounds good to me. I'm no expert at this C interface business.

mwlon (approved):

Skielex:
I'll ask on Discord if anyone has a opinion on this. I think moving to modern fixed size types would be the right thing, but I'm no expert on C either.

Skielex:
Not a lot of response. I'll leave the others as they are.

PR #256, opened 2024-12-07
safer delta lookback decoding

mwlon:


mwlon:
Ah you're right, only panics were possible. I misremembered since it had been a while since I merged this. I'll correct the release notes.

PR #283, opened 2025-05-23
test for stack size of different structs

mwlon:
This is mainly to surface and track the information of how damn large some of these are. We could do something better.

@Skielex this may be informative if you're going to look into startup time performance.


Skielex:
I like this idea. Should we include all latest types? Decompressors drop quite significantly in size for smaller latents.

mwlon:
I don't think it's necessary to track more than u64 here. That's the largest latent size, and these are all just linearly related to latent size.

Skielex (approved):
lgtm
I assume we'll look at the todos in another PR

PR #297, opened 2025-06-13
Fix standalone decompression bug in FFI libs when no n_hint is present

mwlon:
Shared decompression logic with the main crate so that it's harder to mess up.

This bug would only appear in standalone and when running on ancient data or data written by the low-level APIs without an n_hint.

The problem was that we only set the dst vector's capacity according to `n_hint`, which is only a hint and not always present, but kept increasing its length and writing to the end, which could cause a segfault:
```rust
  let mut res: Vec<T> = Vec::with_capacity(n_hint);  // reasonable start
  while let MaybeChunkDecompressor::Some(mut chunk_decompressor) = ... {
    let initial_len = res.len();
    let remaining = chunk_decompressor.n();
    unsafe {
      res.set_len(initial_len + remaining); // DANGER
    }
    let progress = chunk_decompressor.decompress(&mut res[initial_len..])?;  // DANGER
 ```

### DIFF
--- a/pco/src/standalone/decompressor.rs
+++ b/pco/src/standalone/decompressor.rs
@@ -165,6 +165,15 @@ impl FileDecompressor {
     };
     Ok(MaybeChunkDecompressor::Some(res))
   }
+
+  pub fn simple_decompress<T: Number>(&self, mut src: &[u8]) -> PcoResult<Vec<T>> {
+    let mut res = Vec::with_capacity(self.n_hint());
+    while let MaybeChunkDecompressor::Some(mut chunk_decompressor) = self.chunk_decompressor(src)? {
+      chunk_decompressor.decompress_remaining_extend(&mut res)?;
+      src = chunk_decompressor.into_src();
+    }
+    Ok(res)
+  }
 }
 
 /// Holds metadata about a chunk and supports decompression.
--- a/pco/src/standalone/decompressor.rs
+++ b/pco/src/standalone/decompressor.rs
@@ -166,6 +166,17 @@ impl FileDecompressor {
     Ok(MaybeChunkDecompressor::Some(res))
   }
 
+  /// Takes in compressed bytes (after the header, at the start of the chunks)
+  /// and returns a vector of numbers.
+  ///
+  /// Will return an error if there are any compatibility, corruption,
+  /// or insufficient data issues.
+  ///
+  /// This function exists (in addition to the [standalone
+  /// functions][crate::standalone]) because the user may want to peek at the
+  /// dtype, allowing them to know which type `<T>` to use here. There is no
+  /// analagous file compressor method because the user always knows the dtype
+  /// during compression.
   pub fn simple_decompress<T: Number>(&self, mut src: &[u8]) -> PcoResult<Vec<T>> {
     let mut res = Vec::with_capacity(self.n_hint());
     while let MaybeChunkDecompressor::Some(mut chunk_decompressor) = self.chunk_decompressor(src)? {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: pco/src/standalone/decompressor.rs [hunk 1]
```diff
@@ -165,6 +165,26 @@ impl FileDecompressor {
...
+  ///
+  /// This function exists (in addition to the [standalone
+  /// functions][crate::standalone]) because the user may want to peek at the
+  /// dtype, allowing them to know which type `<T>` to use here. There is no
+  /// analagous file compressor method because the user always knows the dtype
+  /// during compression.
+  pub fn simple_decompress<T: Number>(&self, mut src: &[u8]) -> PcoResult<Vec<T>> {
+    let mut res = Vec::with_capacity(self.n_hint());
+    while let MaybeChunkDecompressor::Some(mut chunk_decompressor) = self.chunk_decompressor(src)? {
+      chunk_decompressor.decompress_remaining_extend(&mut res)?;
```
bzinberg, line 183: IIUC, this line (via [this](https://github.com/pcodec/pcodec/blob/c4c747784440ff05b9ab3a9a17fdaa5a821bbdf0/pco/src/standalone/decompressor.rs#L233) line, via [`Vec::reserve`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve)) is where maybe-reallocation happens.  (It needed to happen somewhere, per the bug described in PR description.)

bzinberg: Tangentially, can these 6 lines be replaced with a call to `Vec::resize_with`?
https://github.com/pcodec/pcodec/blob/c4c747784440ff05b9ab3a9a17fdaa5a821bbdf0/pco/src/standalone/decompressor.rs#L231-L236

pco_java/src/main/rust/src/lib.rs
```diff
@@ -68,24 +70,11 @@ fn simple_compress_inner(
...
-    let initial_len = res.len(); // probably always zero to start, since we just created res
-    let remaining = chunk_decompressor.n();
-    unsafe {
-      res.set_len(initial_len + remaining);
-    }
-    let progress = chunk_decompressor.decompress(&mut res[initial_len..])?;
-    assert!(progress.finished);
-    src = chunk_decompressor.into_src();
-  }
-  let num_array = num_array::to_java(env, &res)?;
```
bzinberg, line 88: I see that this piece is a diverged almost-duplicate of the logic in `simple_decompress`, with this copy not having the bounds check.  I'll return to this in a bit to try to understand in detail what else exactly has been deleted here.

pco/src/standalone/decompressor.rs
mwlon: Unfortunately not, since we want to leave the memory uninitialized for performance reasons. We have to do something unsafe here. It would be nice if Rust had an unsafe one-liner for this, but I don't see such a thing.

bzinberg:
I'm attempting to add the Python test to a fresh fork of `main` and see it fail for the expected reason.  I don't see instructions for how to run the Python tests, @mwlon do those exist?

What I did so far:
```shell
cd pco_python
cargo build
python3 -m venv my_venv
source my_venv/bin/activate
pip3 install -e .
pip3 install -r test_requirements.txt
pytest test/test_standalone.py
```
I get an error saying unable to import `pcodec`, despite the fact that I can import pcodec from a Python shell.  Possibly I'm just missing some parameter to the `pytest` invocation telling it to stay inside the virtualenv.  Will return to this in a bit to debug.

mwlon:
@bzinberg there are some instructions in https://github.com/pcodec/pcodec/blob/main/docs%2FCONTRIBUTING.md

bzinberg:
Ok great, for posterity added tag [`pr297-failing-regtest`](https://github.com/pcodec/pcodec/releases/tag/pr297-failing-regtest) pointing to a (branchless) commit where the newly-added Python test (without the rest of the code in this PR) does produce a segfault as expected.

PR #299, opened 2025-06-15
Smaller `Node` struct and replaced while-loop

Skielex:
The reduced size of `Node` appears to improve decode performance by 3-4% on some systems. This is likely due to fewer cache misses.

Replacing the while-loop in `Decoder::new` shouldn't really affect performance, but thought it was nicer to calculate the values.

### DIFF
--- a/pco/src/ans/decoding.rs
+++ b/pco/src/ans/decoding.rs
@@ -3,11 +3,11 @@ use crate::ans::{AnsState, Symbol};
 use crate::constants::Bitlen;
 
 #[derive(Clone, Debug)]
-#[repr(align(16))]
+#[repr(align(8))]
 pub struct Node {
-  pub symbol: Symbol,
-  pub next_state_idx_base: AnsState,
-  pub bits_to_read: Bitlen,
+  pub symbol: u16,
+  pub next_state_idx_base: u16,
+  pub bits_to_read: u16,
 }
 
 #[derive(Clone, Debug)]
@@ -29,8 +29,8 @@ impl Decoder {
         bits_to_read += 1;
       }
       nodes.push(Node {
-        symbol,
-        next_state_idx_base: next_state_base - table_size as AnsState,
+        symbol: symbol as u16,
+        next_state_idx_base: (next_state_base - table_size as AnsState) as u16,
         bits_to_read,
       });
       symbol_x_s[symbol as usize] += 1;
--- a/pco/src/ans/decoding.rs
+++ b/pco/src/ans/decoding.rs
@@ -1,11 +1,11 @@
 use crate::ans::spec::Spec;
-use crate::ans::{AnsState, CompactAnsState, CompactSymbol};
+use crate::ans::{AnsState, CompactAnsState, Symbol};
 use crate::constants::CompactBitlen;
 
 #[derive(Clone, Debug)]
 #[repr(align(8))]
 pub struct Node {
-  pub symbol: CompactSymbol,
+  pub symbol: Symbol,
   pub next_state_idx_base: CompactAnsState,
   pub bits_to_read: CompactBitlen,
 }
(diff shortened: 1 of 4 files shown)


pco/src/ans/decoding.rs
```diff
@@ -1,13 +1,15 @@
...
-use crate::constants::Bitlen;
+use crate::ans::{AnsState, CompactAnsState, Symbol};
+use crate::constants::CompactBitlen;
 
+// Using smallar types for AnsState and Bitlen to reduce the memory footprint
+// of Node. This improves performance, likely due to fewer cache misses.
 #[derive(Clone, Debug)]
-#[repr(align(16))]
+#[repr(align(8))]
 pub struct Node {
```
Skielex, line 9: We could use a smaller `Symbol` size too, but in my tests it seems to hurt performance a little bit, though it may be run-to-run variance.

pco/src/ans/decoding.rs
```diff
@@ -22,16 +24,13 @@ impl Decoder {
...
     let mut symbol_x_s = spec.symbol_weights.clone();
     for &symbol in &spec.state_symbols {
-      let mut next_state_base = symbol_x_s[symbol as usize] as AnsState;
-      let mut bits_to_read = 0;
-      while next_state_base < table_size as AnsState {
-        next_state_base *= 2;
-        bits_to_read += 1;
-      }
+      let next_state_base = symbol_x_s[symbol as usize] as AnsState;
+      let bits_to_read = next_state_base.leading_zeros() - (table_size as AnsState).leading_zeros();
```
Skielex, line 28: Should we test that `next_state_base.leading_zeros() >= (table_size as AnsState).leading_zeros()`?

pco/src/constants.rs
```diff
@@ -2,6 +2,7 @@
 // u32 is performant because it aligns with other things and doesn't require
 // masking
 // exposed in public API
+pub(crate) type CompactBitlen = u16;
```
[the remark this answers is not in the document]
mwlon: Yeah, let's not do this.

mwlon:
Does performance anywhere get worse if we don't mint `Compact*` and just change the type across the board?

pco/src/ans/decoding.rs
Skielex: Yeah, but it might be a fluke. I've changed it now as you suggested.

Skielex:
> Does performance anywhere get worse if we don't mint `Compact*` and just change the type across the board?

Just changing the three types to `u16` tanks performance (like 86ms -> 137ms). It was the first thing I tried a while back:
https://github.com/pcodec/pcodec/compare/main...Skielex:pcodec:16bit-bitlen

I haven't looked into exactly what the issue is, but given how bad it is, I assume the compiler fails to vectorize. But that could perhaps be fixed.

mwlon:
I got out my old x64 laptop and took a look at what's going on. Seems like this PR increases the number of loads, and on x64 for some reason using 32-bit bitlens and 16-bit symbols increases the number of loads even more. It's weird to me that it even does multiple loads per tANS lookup; maybe I should force it to do a single load and some masking/shifting to get the values contained.

mwlon:
Weirdly I also see instructions like `movl %ebx, %ebx`. I don't understand why these would be here

pco/src/latent_page_decompressor.rs
```diff
@@ -14,14 +14,14 @@ use crate::{ans, bit_reader, delta, read_write_uint};
 #[derive(Clone, Copy, Debug)]
 pub struct BinDecompressionInfo<L: Latent> {
   pub lower: L,
-  pub offset_bits: Bitlen,
+  pub offset_bits: CompactBitlen,
```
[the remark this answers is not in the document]
Skielex: Hamm, yeah, reverted too much.
We could leave it `u16` to make `BinDecompressionInfo` smaller for `u16` latents, but it's probably not worth it. Will fix.

mwlon:
lol packing even the loop variables into small portions of larger data types is pretty next level. This appears to be ~4% faster than main on aarch64, which is about 1% faster than the next fastest approach I've seen so far. The code is frankly horrifying though, so I'd like to dig in and figure out which choices here really contribute to that, which ones are hurting us on some CPUs, and where we can find the best compromise. I think I'll have some more time for this tomorrow

PR #302, opened 2025-06-19
fix cargo build on macos;

vectorsss:
Hi, there is an error when I try to build on MacOS with M1 chip.

error: linking with `cc` failed: exit status: 1

According to the documentation of PYO3, I fix the problem successfully.
https://pyo3.rs/v0.20.2/building_and_distribution#macos

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,15 +1,8 @@
 [target.'cfg(target_arch = "x86_64")']
 rustflags = ["-C", "target-feature=+bmi1,+bmi2,+avx2"]
 
-[target.x86_64-apple-darwin]
+[target.'cfg(target_os = "macos")']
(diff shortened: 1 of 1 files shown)


.cargo/config.toml
```diff
@@ -1,2 +1,15 @@
 [target.'cfg(target_arch = "x86_64")']
 rustflags = ["-C", "target-feature=+bmi1,+bmi2,+avx2"]
+
+[target.x86_64-apple-darwin]
```
mwlon, line 4: Can you combine these into one using `target_os`?

vectorsss: Sure, I will merge it! Thank you!

mwlon:
Thanks for the PR!

PR #303, opened 2025-06-20
Packed ANS decoding table

mwlon:
* Made all ANS decoding table types u16 so they fit in an 8-byte word
* Moved offset_bits into the ANS table directly, rather than bin decompression infos. Previously BinInfo<L> could have size of 6 (rounded up to 8?), 8, or 12 bytes, which could make for an awkward access pattern, but now they always pack nicely.


mwlon:
Total instruction count is the same on aarch64, but fewer adds and more ldrhs. Maybe more ldrhs are harmless since they're all waiting on the same cache line. Main:

  25 add
   1 and
   4 bic
   1 cmp
   4 ldp
   9 ldr
   4 ldrh
   4 lsl
   5 lsr
   8 str
   4 strh
   
This PR:

  21 add
   1 and
   4 bic
   1 cmp
   5 ldr
  16 ldrh
   4 lsl
   5 lsr
  12 str

mwlon:
Across the synthetic benchmarks, I measured 3-5% improvement on my big L1 cache aarch64 machine and 4-8% on my weak Intel machine.

PR #312, opened 2025-07-10
Add `lower` in `decompress_offsets` loop instead of separate function

Skielex:
- Eliminates extra function.
- ~2% decreased decoding time in tests on X86.
- Reduces binary size by a couple of kb.

### DIFF
--- a/pco/src/latent_page_decompressor.rs
+++ b/pco/src/latent_page_decompressor.rs
@@ -140,32 +140,25 @@ impl<L: Latent> LatentPageDecompressor<L> {
     let base_bit_idx = reader.bit_idx();
     let src = reader.src;
     let state = &mut self.state;
-    for (dst, (&offset_bits, &offset_bits_csum)) in dst.iter_mut().zip(
-      state
-        .offset_bits_scratch
-        .iter()
-        .zip(state.offset_bits_csum_scratch.iter()),
+    for (dst, (&offset_bits, (&offset_bits_csum, &lower))) in dst.iter_mut().zip(
+      state.offset_bits_scratch.iter().zip(
+        state
+          .offset_bits_csum_scratch
+          .iter()
+          .zip(state.lowers_scratch.iter()),
+      ),
     ) {
       let bit_idx = base_bit_idx + offset_bits_csum as usize;
       let byte_idx = bit_idx / 8;
       let bits_past_byte = bit_idx as Bitlen % 8;
-      *dst = bit_reader::read_uint_at::<L, MAX_U64S>(src, byte_idx, bits_past_byte, offset_bits);
+      *dst = bit_reader::read_uint_at::<L, MAX_U64S>(src, byte_idx, bits_past_byte, offset_bits)
+        .wrapping_add(lower);
     }
     let final_bit_idx = base_bit_idx
       + state.offset_bits_csum_scratch[dst.len() - 1] as usize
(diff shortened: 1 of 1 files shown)


Skielex:
Let me know if this changes anything on ARM.

mwlon:
It looks like on main, 3/4 of these loops get compiled to SIMD (I assume 64-bit, MAX_U64S=2 is the non-SIMD one), but on this branch only 1/4 does. It's a few % slower. Investigating.

Skielex:
I don't understand why a simple addition before storing to `dst` would change the vectorization. It just needs to add a SIMD add at the end.

Skielex:
So I tried a lot of different things. The root issue here is that LLVM won't vectorize on `aarch64` if we have three inputs (`offset_bits`, `offset_bits_csum` and `lower`) and one output (`dst`). On `x86` vectorization works fine in this case.

One workaround that keeps vectorization `aarch64` with `wrapping_add` is to merge the two `u32` offset values into a single `u64` and extract them manually like this https://github.com/pcodec/pcodec/compare/main...Skielex:pcodec:aligned+lowers
However, this only appear give about 1-3% improvement on `x86`.

A few other things i noticed:
1. `decompress_offsets` does appear to vectorize for `aarch64` for `u16` latents.
2. AVX-512 (at least on the slightly older Xeons i tried on) tanks performance. So compiling with `native` target is a bad idea. Should we warn people?
3. We're running multiple interations over data in `dst`. If `dst` is poorly aligned this could hurt performance.

mwlon:
Interesting! I wonder if it's worth filling an LLVM issue. I'll take a look tomorrow; might still be worthwhile to take that packed u64 improvement if it applies everywhere.

1. Yep, it's strange.
2. Maybe. I also noticed this on my old laptop. This business can get messy though. Some machines technically have 512 support but do it badly (low throughput or few retirements per cycle); others genuinely support it well. It would be nice to evaluate on some fancy new x64 cpu if you haven't already. If this holds universally, it would be nice to understand why. A plausible culprit is that clock speed reduces when entering avx512 instructions, and that transition causes a delay on the CPU. We could add another compile-time warning if that's the issue.
3. I think multiple passes and misalignment are orthogonal issues, right? We can evaluate fixes to each independently.

Skielex:
Been quite busy recently, so haven't had time to push this forward. Don't think I'll be able to spend much time on this for some time, so I'm considering abandoning this PR.

I do think we're leaving some performance on the table here due to:
- Multiple passes (decompress offsets, add lowers, convert latents to numbers). 
- Poor vectorization of offset decompression for 16-bit (and 8-bit) types.

However, all the solutions I've tried so far seem to provide benefits that outweigh the costs.

As mentioned there could also be penalties from working directly on the `dst` array, if it's poorly aligned, although this is not confirmed. This reason I mention it here is that the penalty should depend on the number of passes we do on the `dst`. But I agree we should probably explore this separately.

I think we should consider explicitly aligning the scratch arrays to 32 or 64 bits. It may not have any impact right now, but it could protects us from future pitfalls, where changing a struct for whatever reason results in poor alignment.

What do you think?  Should we let this be for now and perhaps revisit later? Perhaps create a separate PR for alignment of scratch?

mwlon:
Totally agree. I'll close this for now, and you could just create an issue linking to this discussion. If alignment alone improves performance in some cases (user-provided misaligned dst), then that could be a worthwhile PR by itself.

PR #314, opened 2025-07-17
Faster low compression levels

mwlon:
On my machine at level 0 this was about a 50-60% speedup on the lomax05 dataset. 2/3 of these shortcuts only really apply to level 0, but the 3rd one is still a slight speedup at other low compression levels, e.g. ~4% at level 2.

### DIFF
--- a/pco/src/compression_table.rs
+++ b/pco/src/compression_table.rs
@@ -30,3 +30,9 @@ impl<L: Latent> From<Vec<BinCompressionInfo<L>>> for CompressionTable<L> {
     }
   }
 }
+
+impl<L: Latent> CompressionTable<L> {
+  pub fn is_trivial(&self) -> bool {
+    self.infos.len() <= 1
+  }
+}
--- a/pco/src/latent_batch_dissector.rs
+++ b/pco/src/latent_batch_dissector.rs
@@ -20,6 +20,8 @@ pub struct LatentBatchDissector<'a, L: Latent> {
 
 impl<'a, L: Latent> LatentBatchDissector<'a, L> {
   pub fn new(table: &'a CompressionTable<L>, encoder: &'a ans::Encoder) -> Self {
+    // We initialize the scratch buffer for bin lowers carefully to enable
+    // a shortcut when there's only one bin.
     let default_lower = table
       .infos
       .first()
@@ -100,7 +102,8 @@ impl<'a, L: Latent> LatentBatchDissector<'a, L> {
     ans_final_states: &mut [AnsState; ANS_INTERLEAVING],
   ) {
     if self.encoder.size_log() == 0 {
-      // trivial case: there's only one symbol
+      // trivial case: there's only one symbol. ANS values and states don't
+      // matter.
       ans_bits.fill(0);
       return;
     }
--- a/pco/src/latent_batch_dissector.rs
+++ b/pco/src/latent_batch_dissector.rs
@@ -20,10 +20,15 @@ pub struct LatentBatchDissector<'a, L: Latent> {
 
 impl<'a, L: Latent> LatentBatchDissector<'a, L> {
   pub fn new(table: &'a CompressionTable<L>, encoder: &'a ans::Encoder) -> Self {
+    let default_lower = table
+      .infos
+      .first()
+      .map(|info| info.lower)
+      .unwrap_or(L::ZERO);
     Self {
       table,
       encoder,
-      lower_scratch: [L::ZERO; FULL_BATCH_N],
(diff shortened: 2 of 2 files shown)


mwlon:
@bzinberg This is just a speedup, so there's nothing new to test.

mwlon:
@bzinberg This is just a speedup, so there's nothing new to test. There are existing tests in recovery.rs that use a single bin, which hits all the new codepaths.

pco/src/latent_batch_dissector.rs
```diff
@@ -42,16 +49,33 @@ impl<'a, L: Latent> LatentBatchDissector<'a, L> {
...
     }
 
-    let max_idx = self.table.infos.len().saturating_sub(1);
-    search_idxs
-      .iter_mut()
-      .for_each(|search_idx| *search_idx = min(*search_idx, max_idx));
+    let n_bins = self.table.infos.len();
+    if n_bins < 1 << self.table.search_size_log {
+      // We worked with a balanced binary tree with missing leaves filled, so it
+      // might have overshot some bin indices.
```
bzinberg, line 55: Was this condition previously only hit in the case where `self.table.infos.len()` is either 0 or 1?  From this description it sounds like it could have happened more generally?

pco/src/latent_batch_dissector.rs
```diff
@@ -20,10 +20,17 @@ pub struct LatentBatchDissector<'a, L: Latent> {
 
 impl<'a, L: Latent> LatentBatchDissector<'a, L> {
   pub fn new(table: &'a CompressionTable<L>, encoder: &'a ans::Encoder) -> Self {
+    // We initialize the scratch buffer for bin lowers carefully to enable
+    // a shortcut when there's only one bin.
```
bzinberg, line 24: A tricky thing about changes like this is that the places in the codebase that have to know about this shortcut are not all right next to each other.  Is there a spec/docstring to which we can add a representation invariant along the lines of "If there are 0 or 1 bins, then the scratch buffer `lower_scratch` is always populated.  If there are more than one bin, then `lower_scratch` is not correctly populated until such-and-such methods have been called"?  It might not be that clear-cut, but I wanted to raise the question in case there is a way.

pco/src/latent_batch_dissector.rs
```diff
@@ -42,16 +49,33 @@ impl<'a, L: Latent> LatentBatchDissector<'a, L> {
...
+    }
 
     search_idxs
   }
 
   #[inline(never)]
   fn dissect_bins(&mut self, search_idxs: &[usize], dst_offset_bits: &mut [Bitlen]) {
+    if self.table.is_trivial() {
+      // trivial case: there's at most one bin. We've prepopulated the scratch
+      // buffers with the correct values in this case.
```
bzinberg, line 68: Both scratch buffers?  This might be as simple as "if there is only one symbol, then the symbol is 0," but just in case I'm misunderstanding, as only the `lower_scratch` buffer is mentioned above.

bzinberg (commented):
Added more detailed questions/comments for the sake of either helping us clarify the code or educating me :-)

Will now look at the unit tests.

pco/src/latent_batch_dissector.rs
mwlon: I'm not sure how that would be done with code without some annoying complexity. These functions aren't user-facing, so there are no doc comments, but hopefully these regular comments are sufficient?

mwlon:
> I see these lines in test_edge_cases() exercise the cases where the input has 0 or 1 elements (are there others I'm missing?):

It's not about whether the input has 0 or 1 elements, but whether there are 0 or 1 bins, which always happens at compression level 0 (and sometimes for other reasons). On line 248 in recovery.rs there's an existing test that asserts we have exactly 1 bin.

pco/src/latent_batch_dissector.rs
mwlon: Yes, both buffers. I added to the comment above to make this clearer.

bzinberg (approved):

pco/src/compression_table.rs
```diff
@@ -30,3 +30,9 @@ impl<L: Latent> From<Vec<BinCompressionInfo<L>>> for CompressionTable<L> {
     }
   }
 }
+
+impl<L: Latent> CompressionTable<L> {
+  pub fn is_trivial(&self) -> bool {
+    self.infos.len() <= 1
```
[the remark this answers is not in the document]
mwlon: (good question btw)

PR #320, opened 2025-09-07
Fixed docs for python behavior on multidim arrays and added tests

### LINKED ISSUE #319: Non-flat Numpy arrays throw error on compression
A while back I noticed that trying to compress non-flat Numpy arrays no longer works.
```
______________________________________________________________________ test_round_trip_decompress_into[u8-900] ______________________________________________________________________

length = 900, dtype = 'u8'

    @pytest.mark.parametrize("length", all_lengths)
    @pytest.mark.parametrize("dtype", all_dtypes)
    def test_round_trip_decompress_into(length, dtype):
        data = np.random.uniform(0, 1000, size=length).astype(dtype).reshape(-1, 1)
>       compressed = standalone.simple_compress(data, ChunkConfig())
E       TypeError: 'ndarray' object cannot be converted to 'PyArray<T, D>'

pco_python/test/test_standalone.py:20: TypeError
============================================================================== short test summary info ==============================================================================
FAILED pco_python/test/test_standalone.py::test_round_trip_decompress_into[u8-900] - TypeError: 'ndarray' object cannot be converted to 'PyArray<T, D>'
=========================================================================== 18 failed, 39 passed in 0.93s ===========================================================================
```
I don't know if this change was intentional or not. If so, the documentation for the Python function should be updated as it says that "any shape" is allowed:
``` rust
pub fn register(m: &Bound<PyModule>) -> PyResult<()> {
  /// Compresses an array into a standalone format.
  ///
  /// :param nums: numpy array to compress. This may have any shape.
  ///   However, it must be contiguous, and only the following data types are
  ///   supported: float16, float32, float64, int16, int32, int64, uint16, uint32, uint64.
  /// :param config: a ChunkConfig object containing compression level and
  ///   other settings.

mwlon:
Fixes #319

### DIFF
--- a/pco_python/src/standalone.rs
+++ b/pco_python/src/standalone.rs
@@ -67,7 +67,7 @@ pub fn register(m: &Bound<PyModule>) -> PyResult<()> {
     match_number_enum!(
       number_type,
       NumberType<T> => {
-        simple_compress_generic(py, nums.downcast::<PyArray1<T>>()?, &config)
+        simple_compress_generic(py, utils::downcast_to_flat::<T>(nums)?, &config)
       }
     )
   }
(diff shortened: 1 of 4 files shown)


mwlon:
@Skielex

PR #321, opened 2025-09-15
Read 32-bit offsets

Skielex:
Support for reading data 32-bits at a time. This allows faster decompression of shorter latent types due to better SIMD bandwidth usage. On my PC I see around 5-15% faster decompression on 16-bit types.

I've been playing around a bit and this is the solution with the fewest changes that seemed decent. That said, I'm not a big fan of the matches on numbers (of bytes or u64s). Seems like an `enum` would be better suited.

We could also extend this change to be used for `write_uint` as well, but I'm not sure if it'll actually provide a performance gain. If we do that, I definitely think we should change from `usize` to `enum`.

One thing to consider is that avoiding `u64` intermediate representation likely only improves performance for non-`u64` latents. So even if `bytes_per_offset < 25` for a 64-bit type, because the output latents are 64 bits wide, it's not clear that reading 32 bit bites will improve throughput. Changes in the compiled SIMD could have the opposite effect. We should benchmark this.

This change does not appear to fix missing vectorization of `decompress_offsets` for `u16` latents on `aarch64`.

### DIFF
--- a/pco/src/bit_reader.rs
+++ b/pco/src/bit_reader.rs
@@ -18,20 +18,30 @@ pub unsafe fn u64_at(src: &[u8], byte_idx: usize) -> u64 {
   u64::from_le_bytes(raw_bytes)
 }
 
+// Q: Why is there also a u32 version?
+// A: This allows for better use of SIMD bandwidth when reading smaller latent
+//    types compared to the u64 version.
 #[inline]
-pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_U64S: usize>(
+pub unsafe fn u32_at(src: &[u8], byte_idx: usize) -> u32 {
+  let raw_bytes = *(src.as_ptr().add(byte_idx) as *const [u8; 4]);
+  u32::from_le_bytes(raw_bytes)
+}
+
+#[inline]
+pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_BYTES: usize>(
   src: &[u8],
   mut byte_idx: usize,
   bits_past_byte: Bitlen,
   n: Bitlen,
 ) -> U {
   // Q: Why is this fast?
-  // A: The 0..MAX_U64S can be unrolled at compile time, allowing really
-  // fast SIMD stuff at least in the MAX_U64S=1 (most common) case.
+  // A: The 0..MAX_BYTES can be unrolled at compile time, allowing really
+  // fast SIMD stuff at least in the MAX_BYTES<=8 (most common) case.
   //
   // Q: Why does this work?
-  // A: We set MAX_U64S so that,
-  //    0  to 57  bit reads -> 1 u64
+  // A: We set MAX_BYTES so that,
+  //    0  to 25  bit reads -> 1 u32
+  //    25 to 57  bit reads -> 1 u64
   //    58 to 113 bit reads -> 2 u64's
   //    113 to 128 bit reads -> 3 u64's
   //    During the 1st u64 (prior to the loop), we read all bytes from the
@@ -47,17 +57,23 @@ pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_U64S: usize>(
   //    For the 3rd u64 and onward, we skip 8 bytes forward. Due to how we
   //    handled the 2nd u64, the most we'll ever need to shift by is
   //    U::BITS - 8, which is safe.
-  let mut res = U::from_u64(u64_at(src, byte_idx) >> bits_past_byte);
-  let mut processed = min(n, 56 - bits_past_byte);
-  byte_idx += 7;
 
-  for _ in 0..MAX_U64S - 1 {
-    res |= U::from_u64(u64_at(src, byte_idx)) << processed;
-    processed += 64;
-    byte_idx += 8;
-  }
+  let res = if MAX_BYTES < 8 {
+    U::from_u32(u32_at(src, byte_idx) >> bits_past_byte)
+  } else {
+    let mut res = U::from_u64(u64_at(src, byte_idx) >> bits_past_byte);
+    let mut processed = min(n, 56 - bits_past_byte);
+    byte_idx += 7;
 
-  if MAX_U64S * 64 <= U::BITS as usize {
+    for _ in (0..MAX_BYTES - 8).step_by(8) {
+      res |= U::from_u64(u64_at(src, byte_idx)) << processed;
+      processed += 64;
+      byte_idx += 8;
+    }
+    res
+  };
+
+  if MAX_BYTES * 8 <= U::BITS as usize {
     bits::lowest_bits_fast(res, n)
   } else {
     bits::lowest_bits(res, n)
@@ -125,20 +141,38 @@ impl<'a> BitReader<'a> {
 
   pub unsafe fn read_uint<U: ReadWriteUint>(&mut self, n: Bitlen) -> U {
     self.refill();
-    let res = match U::MAX_U64S {
-      1 => read_uint_at::<U, 1>(
+    let res = match U::MAX_BYTES {
+      4 => {
+        if U::BITS <= 32 {
+          // only decompress 4-byte offsets if the latent type is 32 bits or smaller
+          read_uint_at::<U, 4>(
+            self.src,
+            self.stale_byte_idx,
+            self.bits_past_byte,
+            n,
+          )
+        } else {
+          read_uint_at::<U, 8>(
+            self.src,
+            self.stale_byte_idx,
+            self.bits_past_byte,
+            n,
+          )
+        }
+      },
+      8 => read_uint_at::<U, 8>(
         self.src,
         self.stale_byte_idx,
         self.bits_past_byte,
         n,
       ),
-      2 => read_uint_at::<U, 2>(
+      16 => read_uint_at::<U, 16>(
         self.src,
         self.stale_byte_idx,
         self.bits_past_byte,
         n,
       ),
-      3 => read_uint_at::<U, 3>(
+      24 => read_uint_at::<U, 24>(
         self.src,
         self.stale_byte_idx,
         self.bits_past_byte,
@@ -146,8 +180,8 @@ impl<'a> BitReader<'a> {
       ),
       0 => panic!("[BitReader] data type cannot have 0 bits"),
       _ => panic!(
-        "[BitReader] data type too large (extra u64's {} > 2)",
-        U::MAX_U64S
+        "[BitReader] data type too large (bytes {} > 24)",
+        U::MAX_BYTES
       ),
     };
     self.consume(n);
--- a/pco/src/data_types/mod.rs
+++ b/pco/src/data_types/mod.rs
@@ -119,6 +119,10 @@ pub trait Latent:
 
   /// Converts a `usize` into this type. Panics if the conversion is
   /// impossible.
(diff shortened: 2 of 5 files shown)


pco/src/latent_page_decompressor.rs
```diff
@@ -153,7 +153,7 @@ impl<L: Latent> LatentPageDecompressor<L> {
   }
 
   #[inline(never)]
-  unsafe fn decompress_offsets<const MAX_U64S: usize>(
+  unsafe fn decompress_offsets<const MAX_BYTES: usize>(
```
mwlon, line 156: What if we make an enum like
```
#[repr(u8)]
enum ReadWidth {
  U32,
  U64,
  U64x2,
}
```
and generally match on this thing, but use like `decompress_offsets::<read_width as u8>(...)` when it's needed in generic parameters?

pco/src/bit_reader.rs
```diff
@@ -125,29 +141,47 @@ impl<'a> BitReader<'a> {
 
   pub unsafe fn read_uint<U: ReadWriteUint>(&mut self, n: Bitlen) -> U {
     self.refill();
-    let res = match U::MAX_U64S {
-      1 => read_uint_at::<U, 1>(
+    let res = match U::MAX_BYTES {
+      4 => {
+        if U::BITS <= 32 {
```
mwlon, line 146: Does this actually improve performance? I wonder if we can always use 32 bit reads here and let the compiler deal with it.

pco/src/bit_reader.rs
```diff
@@ -18,20 +18,30 @@ pub unsafe fn u64_at(src: &[u8], byte_idx: usize) -> u64 {
...
+//    types compared to the u64 version.
 #[inline]
-pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_U64S: usize>(
+pub unsafe fn u32_at(src: &[u8], byte_idx: usize) -> u32 {
+  let raw_bytes = *(src.as_ptr().add(byte_idx) as *const [u8; 4]);
+  u32::from_le_bytes(raw_bytes)
+}
+
+#[inline]
+pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_BYTES: usize>(
```
mwlon, line 31: Now that we might use u32s, I don't like the casework I'm doing here anymore. How about we just make 3 functions for u32, u64, and u64x2; skip u64x3 for now since we don't have 128 bit latents (yet); and route between these 3 in `read_uint_at`?

Skielex: I tried doing individual functions, but I didn't like the results. They were either a single line or repeated the same code. I think your original solution is quite elegant as it support extending to 2+ `u64` in the future without changing `read_uint_at`.

Anyway I changed the code a little to something a bit nicer imho.

pco/src/bit_reader.rs
Skielex: I'm gonna do some proper benchmarks I think.

Skielex: Managed to get some work done on my benchmark setup. I did 30 runs on each dataset on `main` and `short-reads`. For each bench input I picked the median of the 30 runs and divided `short-reads` medians by `main` medians. White is less than 1% change, green is decrease in time, and is and red is increase.

Grouped by dataset and dtype:
image

Grouped by dtype:
image

Results look good for smaller types. For 64-bit types it's a mixed bag, but overall we seem to lose some performance for 64-bit types, even though some of the 64-bit inputs like `i64_KafkaOffset` and `f64_slow_cosine` see a 10% improvement.

All benchmarks were run on a Xeon Gold 6226 HPC node and pinned to a specific core.

mwlon: Thanks, that's pretty handy.

Outside of this PR: do you think we should add some functionality like this to the bench command? Could allow multiple inputs and add options for mean and quantile aggregations

Skielex: > Outside of this PR: do you think we should add some functionality like this to the bench command? Could allow multiple inputs and add options for mean and quantile aggregations

I think what I've done works better as scripts. I think we should keep the Rust bench code simple. Let's discuss this somewhere else.

> It's curious that anything would get slower. I might take a look at the assembly to see what changed, but this looks like a worthy change even if we can't improve it.
The main difference seems to be that `u32` reads for `u64` latents uses `vpinsrd` instead of the `vmovq` used for loading `u64`. At a higher level I think it makes sense that reading data is the same type as the output could be an advantage.

In any case, I did some benchmark to confirm on on the last commit (cada58d7bf4a1857b1a68b7ff6090ed8012a9c14) and the results seem to confirm that avoiding we shouldn't read using a smaller type than the latent.

image
image

This clearly improves 64-bit results compared to 0f05ee33a3e59817a363e6a2025f04d580dd49b5.

Skielex: What I'm struggling to understand here is the 10% improvements for a few of the 64-bit types.

pco/src/bit_reader.rs
```diff
@@ -47,23 +57,52 @@ pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_U64S: usize>(
...
-  let mut processed = min(n, 56 - bits_past_byte);
-  byte_idx += 7;
 
-  for _ in 0..MAX_U64S - 1 {
-    res |= U::from_u64(u64_at(src, byte_idx)) << processed;
-    processed += 64;
-    byte_idx += 8;
-  }
+  let res = match MAX_BYTES {
+    4 => {
```
mwlon, line 62: Could you use a match guard to flatten this logic? https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html

pco/src/read_write_uint.rs
```diff
@@ -4,6 +4,22 @@ use std::ops::{Add, BitAnd, BitAndAssign, BitOr, BitOrAssign, Shl, Shr, Sub};
...
+    8
+  } else if precision <= 113 {
+    16
+  } else {
+    24
+  }
+}
+
 // this applies to reading and also works for byte-aligned precisions
 pub const fn calc_max_u64s(precision: Bitlen) -> usize {
```
mwlon, line 24: What do we still need this one for? Can we remove it?

Skielex: We can, but then we have to change `write_uint`. If we do that, we should probably also change `calc_max_u64s_for_writing`, which would require changing `write_dissected_batch`. 

I actually changed all this in my first go, but reversed to keep changes confined. What would you prefer?

mwlon:
I tried it out on aarch64. There indeed seems to be some speedup, but much smaller, like only 1% faster for the air quality dataset.

pco/src/bit_reader.rs
```diff
@@ -47,23 +57,44 @@ pub unsafe fn read_uint_at<U: ReadWriteUint, const MAX_U64S: usize>(
   //    For the 3rd u64 and onward, we skip 8 bytes forward. Due to how we
   //    handled the 2nd u64, the most we'll ever need to shift by is
   //    U::BITS - 8, which is safe.
```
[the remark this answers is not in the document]
Skielex: I'm considering changing `MAX_BYTES` to `READ_BYTES` or `READ_WIDTH`.

pco/src/read_write_uint.rs
mwlon: Let's get rid of it and just use the `calc_max_u64s_for_writing` function;

* `calc_max_u64s` is now only used in `BitWriter::write_uint`, which uses the max type size, which is never going to be on the meaningful threshold of 56/57
* the performance intensive bitpacking during compression all happens through `LatentChunkCompressor::write_dissected_batch`, which is already using `calc_max_u64s_for_writing` instead

mwlon: I just tested it, and removing it causes no harm to performance on my machine

pco/src/bit_reader.rs
```diff
@@ -18,52 +18,80 @@ pub unsafe fn u64_at(src: &[u8], byte_idx: usize) -> u64 {
...
+}
+
+#[inline]
+unsafe fn read_u64x2_at<U: ReadWriteUint>(
+  src: &[u8],
+  byte_idx: usize,
+  bits_past_byte: Bitlen,
+  n: Bitlen,
+) -> U {
+  let res = U::from_u64(u64_at(src, byte_idx) >> bits_past_byte);
```
mwlon, line 90: Can we tweak this a bit more?

* We can actually just use `processed = 56 - bits_past_byte;`, so this function doesn't need `n` at all. I think I could have just done this the whole time?
* I try to only use `res` for the thing to return from the function, so can we call it something more meaningful like `first_word`?

pco/src/bit_reader.rs
```diff
@@ -125,29 +153,29 @@ impl<'a> BitReader<'a> {
...
         self.src,
         self.stale_byte_idx,
         self.bits_past_byte,
         n,
       ),
-      0 => panic!("[BitReader] data type cannot have 0 bits"),
-      _ => panic!(
-        "[BitReader] data type too large (extra u64's {} > 2)",
-        U::MAX_U64S
+      0 => unreachable!("[BitReader] data type cannot have 0 bits"),
```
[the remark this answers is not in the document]
Skielex: Do you want me to update `write_uint` to align with the updated `read_uint`?

pco/src/bit_reader.rs
Skielex: Yeah, was actually wondering why we needed `min(n, ...)`.

pco/src/data_types/mod.rs
```diff
@@ -117,6 +117,10 @@ pub trait Latent:
   const MAX: Self;
   const BITS: Bitlen;
 
+  /// Converts a `usize` into this type. Panics if the conversion is
```
mwlon, line 120: Just noticed the comments for this and from_u64 are both wrong and say usize

mwlon (approved):

Skielex:
Just realized we might be able to utilized `lower_bits` in more cases. Right now, if we read a `u32` latent using a 8 byte read, we end up using `lower_bits`, not `lower_bits_fast`. This shouldn't be necessary. 

I'll have a look a this when I'm at my PC.

mwlon:
@Skielex I think you're onto something good here. What if instead of `READ_BYTES` we do `MAX_READ_BITS` or something, with values in `{0, 25, 57, 113}`? Then it would be a bit more clear whether to use `lowest_bits_fast` or not. It still wouldn't be optimal for say 31 bit reads, but we wouldn't be too far off, and it would be fairly simple.

Skielex:
Pushed my current code. Last bench results look like this:
image

pco/src/bit_reader.rs
```diff
@@ -18,50 +17,95 @@ pub unsafe fn u64_at(src: &[u8], byte_idx: usize) -> u64 {
...
+  byte_idx: usize,
+  bits_past_byte: Bitlen,
+  n: Bitlen,
+) -> U {
+  debug_assert!(n <= 113);
+  let first_word = u64_at(src, byte_idx) >> bits_past_byte;
+  let processed = 56 - bits_past_byte;
+  let second_word = u64_at(src, byte_idx + 7) << processed;
+  U::from_u64(bits::lowest_bits(
+    first_word | second_word,
```
mwlon, line 106: I know it doesn't affect the current set of latent types, but this won't work if we try adding `u128`

PR #324, opened 2025-09-20
Update vortex dependency

mwlon:
to fix a bug (related to incomplete or misordered buffers perhaps?) that affected versions through 0.52.1


mwlon:
waiting on this until they release 0.53

PR #334, opened 2025-11-10
Scratch space for compression on all pages in a chunk

mwlon:
This improves compression speed slightly when using many small pages. The most extreme case is about 20% speedup on constant data with pages of size 256.

Also moved all `ChunkLatentDissector` logic into `ChunkLatentCompressor` since the former basically only held references to the latter.


Skielex:
This doesn't seem to benefit `standalone` compress with paging since it uses `write_chunk`, which call `self.inner.write_page(0, dst)`. Would it be worth it to reuse the scratch here too? I could image many people will use `standalone` directly, or base their code on it.

PR #335, opened 2025-11-15
Simpler page latent scratch

mwlon:
No more separate scratch just for PageDecompressor when PageLatentDecompressor already has one.

Changed the data flow FROM
* PLDs add offsets and stored lowers into a provided dst array
  * delta and secondary latent variables use an extra scratch buffer; primary uses the user's provided dst directly
* join_latents takes the primary(==dst) and separate secondary latents and joins in place with some wacky transmute

TO
* PLDs add offsets into stored lowers in place
  * all latent variables are implemented in a more consistent way AND we don't need PageDecompressor to hold extra scratch buffers anymore
* join_latents takes the primary and secondary latents and joins them into dst using intuitive operations

Benefits:
* Simpler, easier to understand code
* Faster because fewer memory traversals
* This works around the auto-vectorization limitation on aarch64 when zipping 3 iterables during `decompress_offsets`, meaning we can use the same code for aarch64 and x64.

before vs after:
```
╭──────────────┬───────┬─────────────┬───────────────┬─────────────────╮
│ dataset      │ codec │ compress_dt │ decompress_dt │ compressed_size │
├──────────────┼───────┼─────────────┼───────────────┼─────────────────┤
│ i64_constant │ pco   │         0ns │     709.895µs │              79 │
│ f64_diablo   │ pco   │         0ns │    5.676583ms │         1850158 │
│ i64_lomax05  │ pco   │         0ns │    2.490021ms │         1729841 │
│ <sum>        │ pco   │         0ns │    8.876499ms │         3580078 │
╰──────────────┴───────┴─────────────┴───────────────┴─────────────────╯

╭──────────────┬───────┬─────────────┬───────────────┬─────────────────╮
│ dataset      │ codec │ compress_dt │ decompress_dt │ compressed_size │
├──────────────┼───────┼─────────────┼───────────────┼─────────────────┤
│ i64_constant │ pco   │         0ns │     514.542µs │              79 │
│ f64_diablo   │ pco   │         0ns │    5.261625ms │         1850158 │
│ i64_lomax05  │ pco   │         0ns │    2.250375ms │         1729841 │
│ <sum>        │ pco   │         0ns │    8.026542ms │         3580078 │
╰──────────────┴───────┴─────────────┴───────────────┴─────────────────╯
```

### DIFF
--- a/pco/src/page_latent_decompressor.rs
+++ b/pco/src/page_latent_decompressor.rs
@@ -1,5 +1,9 @@
+use std::any::TypeId;
 use std::fmt::Debug;
 use std::ops::{Deref, DerefMut};
+use std::ptr;
+
+use wide::{u32x4, u64x2, u64x4};
 
 use crate::ans::{AnsState, Spec};
 use crate::bit_reader::BitReader;
@@ -57,6 +61,30 @@ impl<L: Latent> State<L> {
     *self.latents.get_unchecked_mut(i) = lower;
   }
 }
+
+#[inline]
+unsafe fn u64x4_from_slice(p: *const u64) -> u64x4 {
+  u64x4::new(*(p as *const [u64; 4]))
+}
+
+#[inline]
+unsafe fn u64x4_from_u32_slice(slice: &[u32]) -> u64x4 {
+  // println!("A {:?}", &slice[..2]);
+  u64x4::new([
+    slice[0] as u64,
+    slice[1] as u64,
+    slice[2] as u64,
+    slice[3] as u64,
+  ])
+  // let p = slice.as_ptr();
+  // u64x2::new([*p as u64, *(p.add(1)) as u64])
+}
+
+// #[inline]
+// unsafe fn u32x2_from_slice(values: &[u32]) -> u32x2 {
+//   let p = values.as_ptr();
+//   u32x2::new(*(p as *const [u32; 2]))
+// }
 
 #[derive(Clone, Debug)]
 pub struct PageLatentDecompressor<L: Latent> {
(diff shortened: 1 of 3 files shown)


pco/src/page_latent_decompressor.rs
```diff
@@ -156,71 +157,53 @@ impl<L: Latent> PageLatentDecompressor<L> {
...
     let src = reader.src;
     let state = &mut self.state;
-    for (dst, (&offset_bits, (&offset_bits_csum, &lower))) in dst.iter_mut().zip(
-      state.offset_bits_scratch.iter().zip(
-        state
-          .offset_bits_csum_scratch
-          .iter()
-          .zip(state.lowers_scratch.iter()),
-      ),
+    for (&offset_bits, (&offset_bits_csum, latent)) in state.offset_bits_scratch.iter().take(n).zip(
```
Skielex: Looks like assembly size would be reduced from using slicing instead:
``` rust
    for (&offset_bits, (&offset_bits_csum, latent)) in state.offset_bits_scratch[..n].iter().zip(
      state.offset_bits_csum_scratch[..n]
        .iter()
        .zip(state.latents[..n].iter_mut()),
```
This is improved further by moving `assert!(batch_n <= FULL_BATCH_N);` (see other comment).

pco/src/page_latent_decompressor.rs
```diff
@@ -156,71 +157,53 @@ impl<L: Latent> PageLatentDecompressor<L> {
...
-  pub unsafe fn decompress_batch_pre_delta(&mut self, reader: &mut BitReader, dst: &mut [L]) {
-    if dst.is_empty() {
+  pub unsafe fn decompress_batch_pre_delta(&mut self, reader: &mut BitReader, batch_n: usize) {
+    if batch_n == 0 || self.is_constant {
       return;
     }
 
     if self.needs_ans {
-      let batch_n = dst.len();
       assert!(batch_n <= FULL_BATCH_N);
```
Skielex: Move assert outside if-statement to reduce size of `decompress_offsets`.

pco/src/page_latent_decompressor.rs
```diff
@@ -48,7 +49,7 @@ impl<L: Latent> State<L> {
     unsafe {
```
Skielex: Do we actually need unsafe here?

pco/src/lib.rs
```diff
@@ -42,6 +42,7 @@ mod compression_intermediates;
 mod compression_table;
 mod constants;
 mod delta;
+pub mod dyn_latent_slice;
```
Skielex: Is this intended to be public?

PR #336, opened 2025-11-21
Recalculate bytes per offset for batches

Skielex:
Recalculating `bytes_per_offset` for each batch appears to improve performance.

This could probably be fine tuned with a more dynamic approach. For instance, we could look at the frequency of high offsets in the nodes to begin with and use that to determine whether we think it's worth it to recalculate for each batch. Alternatively, we could look at the value of recalculating `bytes_per_offset` for previous batches and stop recalculating if recalculated values are not lower than the upper bound.

Recalculating `bytes_per_offset` should provide the most benefit when:
- The recalculated value is 0, allowing lowers to be used as is.
- The upper bound is >8, requiring the slower 2x64 bit read.

There should also be a small improvement when we can read `u32` instead of `u64`, but I'd guess this would be less than the recalculation time. In other words, recalculating `bytes_per_offset` should be most beneficial when many batches get `bytes_per_offset == 0` or a few batches have `bytes_per_offset > 8`.

### DIFF
--- a/pco/src/page_latent_decompressor.rs
+++ b/pco/src/page_latent_decompressor.rs
@@ -57,6 +57,7 @@ impl<L: Latent> State<L> {
 pub struct PageLatentDecompressor<L: Latent> {
   // known information about this latent variable
   bytes_per_offset: usize,
+  bytes_per_offset_matched_ub: i64,
   state_lowers: Vec<L>,
   needs_ans: bool,
   decoder: ans::Decoder,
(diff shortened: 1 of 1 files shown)


Skielex:
image

pco/src/page_latent_decompressor.rs
```diff
@@ -223,6 +224,42 @@ impl<L: Latent> PageLatentDecompressor<L> {
...
+    // Cases:
+    // - `self.bytes_per_offset == 0`: all offsets are zero, so no need to
+    //   recompute.
+    // - `self.bytes_per_offset_matched_ub >= threshold`: we've seen several
+    //   batches in a row where the computed `bytes_per_offset` matches the
+    //   previous upper bound. This suggests that the upper bound is accurate, so we
+    //   can just use it directly.
+    // - `dst.len() < FULL_BATCH_N`: improves performance for case where there's
+    //   just one batch since we've already calculated the upper bound.
+    const MATCHED_UPPER_BOUND_THRESHOLD_COUNT: i64 = 8;
```
Skielex: Would be nice if there's come clever way to choose this.

Skielex:
An alternative strategy would be to keep checking periodically to see if we should recalculate. Something like: Recalculate `bytes_per_offset` for every Kth batch OR if `bytes_per_offset` < upper bound for previous batch. We could sample at random intervals to avoid being hurt by specific patterns.

mwlon:
This is a cool idea, but it does seem tricky and it's a bit surprising to me that it can help much. Like you need a dataset where `2^-18 < P(bin with many offset bits) < 2^-8` (so a typical chunk has at least one very long offset, but a typical batch does not).

Actually this gave me another interesting idea: what if we incentivized bin optimization to avoid making bins that require >57 (or 25) bits? This could help in cases where `2^-8 < P(bin with many offset bits)`, since at default compression ratio we have up to 2^8 bins. But it would only work when all the unoptimized bins are narrow enough but an optimized one is not (and that seems pretty stringent).

Anyway, I'm kinda reluctant on both your idea and my own, because they noticeably increase code complexity and only improve performance in a small number of cases. Do you have a specific case where something like this would help you?

PR #337, opened 2025-12-03
Simpler page latent scratch (attempt 2)

mwlon:
No more PageLatentDecompressor scratch just for latents, when PageDecompressor already has one.

* Contract between PD and PLD:
  * PD must always provide `dst` of length 256 (this is the only part that changed)
  * PLD always modifies `dst` (as opposed to leaving data in scratch like a previous attempt)
* Benefits:
  * Faster in some cases because fewer memory traversals.
  * Works around the auto-vectorization limitation on aarch64 when zipping 3 iterables during decompress_offsets, improving performance and meaning we can use the same code for aarch64 and x64.
  * Reduced # of scratch buffers.
  * We now control the alignment of the PD scratch buffers, FWIW

before vs after:
```
╭──────────────┬───────┬─────────────┬───────────────┬─────────────────╮
│ dataset      │ codec │ compress_dt │ decompress_dt │ compressed_size │
├──────────────┼───────┼─────────────┼───────────────┼─────────────────┤
│ i64_constant │ pco   │         0ns │     486.375µs │              79 │
│ f64_diablo   │ pco   │         0ns │    3.888604ms │         1850158 │
│ i64_dollars  │ pco   │         0ns │    1.660833ms │          619771 │
│ i64_lomax05  │ pco   │         0ns │    1.656021ms │         1729841 │
│ <sum>        │ pco   │         0ns │    7.691833ms │         4199849 │
╰──────────────┴───────┴─────────────┴───────────────┴─────────────────╯

╭──────────────┬───────┬─────────────┬───────────────┬─────────────────╮
│ dataset      │ codec │ compress_dt │ decompress_dt │ compressed_size │
├──────────────┼───────┼─────────────┼───────────────┼─────────────────┤
│ i64_constant │ pco   │         0ns │      473.25µs │              79 │
│ f64_diablo   │ pco   │         0ns │    3.650333ms │         1850158 │
│ i64_dollars  │ pco   │         0ns │    1.516834ms │          619771 │
│ i64_lomax05  │ pco   │         0ns │    1.511937ms │         1729841 │
│ <sum>        │ pco   │         0ns │    7.152354ms │         4199849 │
╰──────────────┴───────┴─────────────┴───────────────┴─────────────────╯
```

### DIFF
--- a/pco/src/tests/stack_sizes.rs
+++ b/pco/src/tests/stack_sizes.rs
@@ -31,6 +31,6 @@ fn test_stack_sizes() {
   );
   assert_eq!(
     mem::size_of::<PageDecompressor<u64, &[u8]>>(),
-    4416
+    224
   );
 }


Skielex:
Performance here shows a small loss on this Intel Xeon:
image

To be honest I'm struggling a bit to see what we achieve here. Sure, the gains on your Mac look very nice, but the code is much more complicated than before and shows a minor performance loss on the Xeon server I tested. Architecturally, I much prefer your initial proposal (052b85fd359764eb141d823d298c8ef22beae7b9). Even if the regression is slightly bigger on Intel (~2% instead of ~1% loss), I think the code structure makes up for it.

As I understand, this new proposal outperforms 052b85fd359764eb141d823d298c8ef22beae7b9 on `aarch64` in some cases where 052b85fd359764eb141d823d298c8ef22beae7b9 fails to vectorize the offsets loop. What I don't get right now, is how this new macro-based solution solves the vectorization issue, and why this solution couldn't work with the generics-based implementation from 052b85fd359764eb141d823d298c8ef22beae7b9.

PR #348, opened 2025-12-25
Simpler page latent scratch, join latents

mwlon:
Changed the data flow in decompression such that PageLatentDecompressors write to their internal buffers instead of an external dst, and joining latents now works without transmutation hacks. To preserve vectorization of `decompress_offsets` on aarch64 during the change (since we no longer have a spare buffer for `add_lowers`), I needed some trickery.

Pros:
* Simpler to explain and work with; fewer scratch buffers and clearer data flow.
* Slightly less (generally L1 cache) memory traversal.
* Can improve performance on aarch64 due to fusing the `decompress_offsets` and `add_lowers` loops into one.
* Will make it possible to use different types for the primary latent variable, since we no longer need to transmute. This will be helpful for the upcoming `Dict` mode, which wants to use u32 indices for all number types.
* Might make it easier to reuse scratch buffers across pages in later PRs (though this is no longer my main objective).

Cons:
* May hurt performance slightly on some x64 CPUs. The reason isn't obvious, but my previous analysis suggested this is probably because the `decompress_offsets` loop is instruction-bound whereas `join_latents` is memory-bandwidth-bound, so as we remove the transmutation hack, `decompress_offsets` gets no faster whereas `join_latents` gets slower.

Performance on my machine before:
```
taxi: 828ms
air quality: 19.9ms
synthetic: 77.8ms
```

and after:
```
taxi: 722ms
air quality: 19.4ms
synthetic: 72.6ms
```

### DIFF
--- a/pco/src/page_latent_decompressor.rs
+++ b/pco/src/page_latent_decompressor.rs
@@ -237,8 +237,6 @@ impl<'a, L: Latent> PageLatentDecompressor<L> {
 
     assert!(batch_n <= FULL_BATCH_N);
     if cld.n_bins > 1 {
-      assert!(batch_n <= FULL_BATCH_N);
-
       if batch_n == FULL_BATCH_N {
         self.decompress_full_ans_symbols(reader, cld);
       } else {
--- a/pco/src/data_types/mod.rs
+++ b/pco/src/data_types/mod.rs
@@ -219,12 +219,7 @@ pub(crate) fn split_latents_classic<T: Number>(nums: &[T]) -> SplitLatents {
 }
 
 pub(crate) fn join_latents_classic<T: Number>(primary: DynLatentSlice, dst: &mut [T]) {
-  for (&l, dst) in primary
-    .downcast::<T::L>()
-    .unwrap()
-    .iter()
-    .zip(dst.iter_mut())
-  {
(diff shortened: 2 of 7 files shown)


pco/src/page_latent_decompressor.rs
```diff
@@ -172,79 +223,29 @@ impl<'a, L: Latent> PageLatentDecompressor<L> {
...
-      let batch_n = dst.len();
       assert!(batch_n <= FULL_BATCH_N);
 
       if batch_n == FULL_BATCH_N {
         self.decompress_full_ans_symbols(reader, cld);
       } else {
         self.decompress_ans_symbols(reader, cld, batch_n);
       }
+    } else {
+      self.latents[..batch_n].fill(cld.state_lowers[0]);
```
Skielex: If no more work needs to be done after this, perhaps we should just return here to make it clearer?

We can make ` (0, _) => unreachable!(...),` then.

Skielex (commented):
Performance compared to `main`:

Image

Should've probably compared to `v1.0`, but didn't.

pco/src/page_latent_decompressor.rs
mwlon: Oh but it's possible there's a single bin that we do need to decompress offsets for (-ANS, +offsets), and it's also possible there are multiple bins, none of which need offsets (+ANS, -offsets). I think it makes sense to have a separate shortcut for each phase.

PR #373, opened 2026-04-26
Specialized conv1@6 decompress

mwlon:
For Conv1@6, this took decompression time on the i16_audio_ish synthetic dataset from 5.00ms to 3.31ms on my machine. That's still a lot slower than NoOp or even Consecutive@2 though:

```
╭───────────────┬─────────────────────────┬─────────────┬───────────────┬─────────────────╮
│ dataset       │ codec                   │ compress_dt │ decompress_dt │ compressed_size │
├───────────────┼─────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ i16_audio_ish │ pco:delta=NoOp          │         0ns │    1.191979ms │         1545267 │
│ i16_audio_ish │ pco:delta=Consecutive@2 │         0ns │    1.824687ms │          678185 │
│ i16_audio_ish │ pco:delta=Conv1@5       │         0ns │    4.835937ms │          636531 │
│ i16_audio_ish │ pco:delta=Conv1@6       │         0ns │    3.308292ms │          636414 │
```

I tried a lot of different variations:
* Algorithm
  * Decode the current latent and then spread its contributions over the next `ORDER` elements (chose this one)
  * OR add all contributions for the current latent we're decoding at once. Specialization to order on this algorithm offered a similar speedup on ARM, but no speedup on x64.
* State
  * Move a bunch of values between registers at each iteration to track state and write them out (chose this one)
  * OR keep a spare scratch array of `FULL_BATCH_N + 2 * ORDER` and do reads and writes instead to avoid all that shuffling. Turned out to be much slower, probably due to the extra loads/stores in the hot path.
  * OR an intermediate approach where I keep a small scratch array of `2 * ORDER`, write values, to it, and every `ORDER` iterations flush latents <- state <- complete scratch <- incomplete scratch. Turned out to be slightly slower.
* Order
  * 6 (chose this one)
  * 1 was barely faster than 6 with any implementation. This strongly suggests the data dependencies and latency to complete this CPU ops are the bottleneck, and also explains why attempts to minimize register shuffling are futile.
  * 4 was still barely faster
  * 8 is where this approach started to slow down

In any case, this speedup is not fast enough to warrant adding conv1@6 as an auto delta encoding.

### DIFF
--- a/pco/src/chunk_config.rs
+++ b/pco/src/chunk_config.rs
@@ -292,6 +292,14 @@ impl ChunkConfig {
             MAX_CONV1_DELTA_ORDER, order,
           )));
         }
+        if !matches!(
+          latent_type,
+          LatentType::U8 | LatentType::U16 | LatentType::U32
+        ) {
+          return Err(PcoError::invalid_argument(
+            "Conv1 delta encoding is only supported for types with 32 or fewer bits",
+          ));
+        }
       }
     }
 
--- a/pco/src/chunk_config.rs
+++ b/pco/src/chunk_config.rs
@@ -17,21 +17,27 @@ pub enum ModeSpec {
   ///
   /// This works well most of the time, but costs some compression time and can
   /// select a bad mode in adversarial cases.
-  /// At present, this will never consider `Dict` mode.
+  /// At present, this does not consider `Dict` mode.
   #[default]
   Auto,
   /// Only uses `Classic` mode.
   Classic,
   /// Tries using `FloatMult` mode with a given `base`.
   ///
+  /// This can be helpful when most numbers are approximately equal modulo
+  /// `base`.
   /// Only applies to floating-point types.
   TryFloatMult(f64),
   /// Tries using `FloatQuant` mode with `k` bits of quantization.
   ///
+  /// This can be helpful when the last `k` bits of the numbers are usually the
+  /// same.
   /// Only applies to floating-point types.
   TryFloatQuant(Bitlen),
   /// Tries using `IntMult` mode with a given `base`.
   ///
+  /// This can be helpful when most numbers are approximately equal modulo
+  /// `base`.
   /// Only applies to integer types.
   TryIntMult(u64),
   /// Tries using `Dict` mode.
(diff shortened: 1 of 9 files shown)


Skielex:
I feel like this could very a lot depending on CPU architecture. We should probably do some testing on various X86 before including this.

Any idea why `Conv1@6` in particular is faster on Mac?

Skielex:
Here's on my local AMD system:
```
cargo run --release -- bench -i data/binary/i16_audio_ish.bin --input-format binary --codecs pco:delta=NoOp,pco:delta=Consecutive@2,pco:delta=Conv1@5,pco:delta=Conv1@6 --iters 500 --no-compress

╭───────────────┬─────────────────────────┬─────────────┬───────────────┬─────────────────╮
│ dataset       │ codec                   │ compress_dt │ decompress_dt │ compressed_size │
├───────────────┼─────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ i16_audio_ish │ pco:delta=NoOp          │         0ns │    1.971066ms │         1545267 │
│ i16_audio_ish │ pco:delta=Consecutive@2 │         0ns │    2.419142ms │          678185 │
│ i16_audio_ish │ pco:delta=Conv1@5       │         0ns │    6.837867ms │          636531 │
│ i16_audio_ish │ pco:delta=Conv1@6       │         0ns │    6.534649ms │          636414 │

**Comment by mwlon** [comment]
Yeah I should mark this as a draft. There are also new tests I need to add

**Comment by Skielex** [comment]
On my AMD 5900X the specialized implementation is slower than the default one.

### Default
```
╭───────────────┬─────────────────────────┬─────────────┬───────────────┬─────────────────╮
│ dataset       │ codec                   │ compress_dt │ decompress_dt │ compressed_size │
├───────────────┼─────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ i16_audio_ish │ pco:delta=Conv1@6       │         0ns │    5.657489ms │          636414 │
│ i16_audio_ish │ pco:delta=Conv1@7       │         0ns │    6.396182ms │          637515 │
```

### `decode_in_place_order_6`
```
╭───────────────┬─────────────────────────┬─────────────┬───────────────┬─────────────────╮
│ dataset       │ codec                   │ compress_dt │ decompress_dt │ compressed_size │
├───────────────┼─────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ i16_audio_ish │ pco:delta=Conv1@6       │         0ns │    6.619604ms │          636414 │
│ i16_audio_ish │ pco:delta=Conv1@7       │         0ns │    5.652614ms │          637515 │
```

I don't understand what's going on with `Conv1@7`, which seems to be affected. For some reason it's faster when `decode_in_place_order_6` is included. What's going on?

mwlon:
@Skielex fear not, I care about x64 performance too and will be testing this on my own ancient intel machine before opening this for real review. I don't know why conv1@7 would be faster there... did you control for your machine's temperature :P ?

mwlon:
@Skielex I find this implementation is slightly faster on x64, but it's not by the degree I was hoping for.

mwlon:
@Skielex I think this is ready if you want to try it out now. I measured like 50% speedup on ARM and 20% on x64, which isn't super impressive for a single case. In my mind the biggest question right now is whether this added complexity is even worth it.

PR #380, opened 2026-07-31
Fix five decoder/API defects found by fuzzing and Kani

enomado:
Five decoder/API defects found by fuzzing and Kani, with fixes and regression tests. The harnesses that found them are now split out into #383 (fuzzing) and #382 (Kani), as requested — this PR is just the fixes and tests.

Everything below is reproducible from this branch.

## What's broken

Two of these hit stock release builds. `cargo fuzz` builds at `opt-level=3` **with** debug assertions and overflow checks, which is not what ships, so I re-checked each finding in both modes:

| # | | stock release | with overflow checks |
|---|---|---|---|
| 1 | `n_hint` → `Vec::with_capacity` | **abort / panic** | abort / panic |
| 2 | `TryFloatQuant(k)`, `k >= L::BITS` | clean `InvalidArgument` | **panic** |
| 3 | `debug_assert!` validating wire data | check absent, arithmetic wraps | **panic** |
| 4 | `mult * base` in `int_mult` | wraps (as intended) | **panic** |
| 5 | `window_n_log` → lookback window | **multi-GB allocation** | multi-GB allocation |

### 1. `n_hint` is trusted as an allocation size

`n_hint` is a varint in the standalone header (`decompressor.rs:113`). Nothing ties it to how much data is actually present — it is a hint — yet `simple_decompress` hands it straight to `Vec::with_capacity` (`decompressor.rs:251`).

* `n_hint = 2^40`, **26-byte file** → 8.8 TB allocation → process **abort**. Not a `PcoResult`; the caller cannot catch it.
* `n_hint = usize::MAX`, **34-byte file** → `capacity overflow` panic.
* Reached from **13 bytes** of unstructured input (172 GB allocation attempt).

These files come out of pco's own public API — `FileCompressor::with_n_hint`, no bit-twiddling. `standalone/guarantee.rs:56` writes `with_n_hint(usize::MAX)` in your own test suite, so it is a legal value on the write side that the read side cannot survive.

Fixed by capping the preallocation at `MAX_N_HINT_PREALLOC = 2^30` numbers.

### 2. `FloatQuant`'s `k` is used before it is validated

`data_types/float.rs:117` calls `float_quant::split_latents(nums, k)` and only *then* reaches the mode-validity check. `split_latents` does `(F::L::ONE << k) - F::L::ONE`. `TryFloatQuant(64)` on `f64` therefore shifts out of range before anything rejects it. `k = 63` is already handled correctly, so the boundary is exactly at `L::BITS`.

Fixed by ordering, not by a new check: every mode spec now emits a `Vec<Bid>`, the winner is chosen, `mode_is_valid` runs on it, and only then is `split_fn` invoked. `mode_is_valid` already rejected `k > PRECISION_BITS`.

### 3. `debug_assert!` used to validate untrusted input

`mode/float_quant.rs:26` asserts `m >> k == 0` on a latent read from the file. A **single-byte** patch of a valid 125-byte file panics on decompress in a debug build; in release the check is compiled out and `lowest_k_bits_max - m` wraps instead.

Fixed by moving the assertion to the compressor, where it belongs (it is a statement about what we emit), and making both of the decoder's arithmetic ops explicitly wrapping — corrupt `m` decodes to garbage, never panics, and costs nothing on the hot path.

### 4. `mult * base` should wrap

`mode/int_mult.rs:48` multiplies two wire-supplied latents with a plain `*`, while the addition on the same line is already `wrapping_add`. Fixed with `wrapping_mul` (added to `LatentPriv` alongside the existing `wrapping_add` / `wrapping_sub`).

### 5. The lookback window is unbounded on read

`metadata/delta_encoding.rs:155` reads `window_n_log = 1 + read_bitlen(5)`, so **1..=32**, and validates only `state_n_log <= window_n_log`. `delta/lookback.rs:192` then allocates `max(1 << window_n_log, 256) * 2` latents. The encoder never emits more than `LOOKBACK_MAX_WINDOW_N_LOG = 15` — **the decoder accepts more than the encoder can produce.**

From a valid **165-byte** lookback file:

| `window_n_log` | window buffer (i64) |
|---|---|
| 15 (encoder max) | 512 KiB |
| 24 | 256 MiB |
| 28 | 4 GiB ← what the fuzzer hit |
| 32 (format max) | 64 GiB |

Fixed with a decoder-side bound of its own, `MAX_DELTA_LOOKBACK_WINDOW_N_LOG = log2(MAX_ENTRIES) = 24` — room above what the encoder emits today, and no chunk can hold enough numbers to use a larger window anyway.

## Tests

`pco/src/tests/corruption.rs` is new: `stability.rs` covers truncation, this covers mutation. A single-byte sweep (every bit flip at every offset, over five mode/delta combinations) plus targeted tests for `n_hint`, `FloatQuant` `k`, and the lookback window bound.

**Every one of these was checked against its own fix reverted** — they all fail without the corresponding change. Suite goes 128 → 134 tests, 0.4s → 1.5s.

## One note

\#1 and #5 are remote DoS on a few bytes of input, so a public PR may not be how you'd have wanted them disclosed. Say the word and I'll close this and re-send privately.

---

A 26-byte file that requests 8.8 TB of memory is, technically, the best compression ratio in this repository. You owe me $10.

### DIFF
--- a/pco/src/constants.rs
+++ b/pco/src/constants.rs
@@ -37,11 +37,11 @@ pub const MAX_CONSECUTIVE_DELTA_ORDER: usize = 7;
 pub const MAX_CONV1_DELTA_ORDER: usize = 32;
 pub const MAX_CONV1_DELTA_QUANTIZATION: Bitlen = (1 << BITS_TO_ENCODE_DELTA_CONV_QUANTIZATION) - 1;
 pub const MAX_ENTRIES: usize = 1 << 24;
-// The wire format allows a lookback window_n_log of up to 32, but the window
-// buffer is allocated eagerly on read, so we cap what we accept. This leaves
-// room for encoders to grow beyond LOOKBACK_MAX_WINDOW_N_LOG while keeping the
-// buffer well under a GB: no chunk can hold more than MAX_ENTRIES numbers
-// anyway, so a larger window could never be fully used.
+// The window_n_log field on the wire is wider than any window we are willing to
+// allocate on read, so reads are capped here. This is deliberately looser than
+// what the encoder emits (ENCODING_LOOKBACK_MAX_WINDOW_N_LOG), leaving room for
+// future encoders; a chunk can hold no more than MAX_ENTRIES numbers, so a
+// larger window could never be fully used anyway.
 pub const MAX_DELTA_LOOKBACK_WINDOW_N_LOG: Bitlen = MAX_ENTRIES.ilog2() as Bitlen;
 pub const MAX_SUPPORTED_PRECISION: Bitlen = 128;
 pub const MAX_SUPPORTED_PRECISION_BYTES: usize = (MAX_SUPPORTED_PRECISION / 8) as usize;
--- a/pco/src/constants.rs
+++ b/pco/src/constants.rs
@@ -37,6 +37,12 @@ pub const MAX_CONSECUTIVE_DELTA_ORDER: usize = 7;
 pub const MAX_CONV1_DELTA_ORDER: usize = 32;
 pub const MAX_CONV1_DELTA_QUANTIZATION: Bitlen = (1 << BITS_TO_ENCODE_DELTA_CONV_QUANTIZATION) - 1;
 pub const MAX_ENTRIES: usize = 1 << 24;
+// The wire format allows a lookback window_n_log of up to 32, but the window
+// buffer is allocated eagerly on read, so we cap what we accept. This leaves
+// room for encoders to grow beyond LOOKBACK_MAX_WINDOW_N_LOG while keeping the
+// buffer well under a GB: no chunk can hold more than MAX_ENTRIES numbers
+// anyway, so a larger window could never be fully used.
+pub const MAX_DELTA_LOOKBACK_WINDOW_N_LOG: Bitlen = MAX_ENTRIES.ilog2() as Bitlen;
 pub const MAX_SUPPORTED_PRECISION: Bitlen = 128;
 pub const MAX_SUPPORTED_PRECISION_BYTES: usize = (MAX_SUPPORTED_PRECISION / 8) as usize;
 pub const MULT_REQUIRED_BITS_SAVED_PER_NUM: f64 = 0.5;
--- a/pco/src/data_types/float.rs
+++ b/pco/src/data_types/float.rs
@@ -1,6 +1,5 @@
 use half::f16;
 
-use super::ModeAndLatents;
 use crate::chunk_config::ModeSpec;
 use crate::compression_intermediates::Bid;
 use crate::constants::Bitlen;
@@ -80,67 +79,56 @@ fn filter_sample<F: Float>(num: &F) -> Option<F> {
   None
 }
 
-fn choose_mode_and_split_latents<F: Float>(
-  nums: &[F],
-  chunk_config: &ChunkConfig,
-) -> PcoResult<ModeAndLatents> {
-  match chunk_config.mode_spec {
+fn choose_mode_bids<F: Float>(nums: &[F], chunk_config: &ChunkConfig) -> PcoResult<Vec<Bid<F>>> {
+  let bids = match chunk_config.mode_spec {
     ModeSpec::Auto => {
       // up to 3 bids: classic, float mult, float quant modes
-      let mut bids: Vec<Bid<F>> = vec![];
-      bids.push(Bid {
+      let mut bids: Vec<Bid<F>> = vec![Bid {
         mode: Mode::Classic,
         bits_saved_per_num: 0.0,
         split_fn: Box::new(|nums| classic::split_latents(nums)),
-      });
+      }];
 
       if let Some(sample) = sampling::choose_sample(nums, filter_sample) {
         bids.extend(float_mult::compute_bid(&sample));
         bids.extend(float_quant::compute_bid(&sample));
       }
 
-      let winning_bid = choose_winning_bid(bids);
-      let latents = (winning_bid.split_fn)(nums);
-      Ok((winning_bid.mode, latents))
+      bids
     }
-    ModeSpec::Classic => Ok((Mode::Classic, classic::split_latents(nums))),
+    ModeSpec::Classic => vec![Bid {
+      mode: Mode::Classic,
+      bits_saved_per_num: 0.0,
+      split_fn: Box::new(|nums| classic::split_latents(nums)),
+    }],
     ModeSpec::TryFloatMult(base_f64) => {
       let base = F::from_f64(base_f64);
-      let mode = Mode::float_mult(base);
       let float_mult_config = FloatMultConfig {
         base,
         inv_base: base.inv(),
       };
-      let latents = float_mult::split_latents(nums, float_mult_config);
-      Ok((mode, latents))
+      vec![Bid {
+        mode: Mode::float_mult(base),
+        bits_saved_per_num: 0.0,
+        split_fn: Box::new(move |nums| float_mult::split_latents(nums, float_mult_config)),
+      }]
     }
-    ModeSpec::TryFloatQuant(k) => {
-      // Must be checked before splitting: split_latents shifts by `k`.
-      if k >= F::L::BITS {
-        return Err(PcoError::invalid_argument(format!(
-          "float quantization k of {} must be less than the type's {} bits",
-          k,
-          F::L::BITS,
-        )));
-      }
-      Ok((
-        Mode::FloatQuant(k),
-        float_quant::split_latents(nums, k),
+    ModeSpec::TryFloatQuant(k) => vec![Bid {
+      mode: Mode::FloatQuant(k),
+      bits_saved_per_num: 0.0,
+      // Note that this is only invoked after the mode is validated; k is a
+      // shift width, so splitting with an out-of-range one is UB-adjacent.
+      split_fn: Box::new(move |nums| float_quant::split_latents(nums, k)),
+    }],
+    ModeSpec::TryIntMult(_) => {
+      return Err(PcoError::invalid_argument(
+        "unable to use int mult mode on floats",
       ))
     }
-    ModeSpec::TryIntMult(_) => Err(PcoError::invalid_argument(
-      "unable to use int mult mode on floats",
-    )),
-    ModeSpec::TryDict => dict::configure_and_split_latents(nums),
-  }
-}
+    ModeSpec::TryDict => vec![dict::compute_bid(nums)],
+  };
 
-// one day we might reuse this for int modes
-fn choose_winning_bid<T: Number>(bids: Vec<Bid<T>>) -> Bid<T> {
-  bids
-    .into_iter()
-    .max_by(|bid0, bid1| bid0.bits_saved_per_num.total_cmp(&bid1.bits_saved_per_num))
-    .expect("bids must be nonempty")
+  Ok(bids)
 }
 
 macro_rules! impl_float {
@@ -396,11 +384,8 @@ macro_rules! impl_float_number {
           Mode::IntMult(_) => false,
         }
       }
-      fn choose_mode_and_split_latents(
-        nums: &[Self],
-        config: &ChunkConfig,
-      ) -> PcoResult<ModeAndLatents> {
-        choose_mode_and_split_latents(nums, config)
+      fn choose_mode_bids(nums: &[Self], config: &ChunkConfig) -> PcoResult<Vec<Bid<Self>>> {
+        choose_mode_bids(nums, config)
       }
 
       #[inline]
@@ -463,13 +448,19 @@ impl_float_number!(f16, u16, 1_u16 << 15, 9);
 #[cfg(test)]
 mod tests {
   use super::*;
+  use crate::compression_intermediates::choose_winning_bid;
   use crate::metadata::{DynLatent, DynLatents};
+
+  fn choose_mode<F: Float>(nums: &[F]) -> Mode {
+    let bids = choose_mode_bids(nums, &ChunkConfig::default()).unwrap();
+    choose_winning_bid(bids).mode
+  }
 
   #[test]
   fn test_choose_mult_mode() {
     let base = 1.5;
     let nums = (0..1000).map(|i| (i as f64) * base).collect::<Vec<_>>();
-    let (mode, _) = choose_mode_and_split_latents(&nums, &ChunkConfig::default()).unwrap();
+    let mode = choose_mode(&nums);
     assert_eq!(mode, Mode::float_mult(base));
     assert!(f64::mode_is_valid(&mode))
   }
@@ -525,7 +516,7 @@ mod tests {
     let nums = (0..1000)
       .map(|i| f64::from_bits(lowest_num_bits + (i << k)))
       .collect::<Vec<_>>();
-    let (mode, _) = choose_mode_and_split_latents(&nums, &ChunkConfig::default()).unwrap();
+    let mode = choose_mode(&nums);
     assert_eq!(mode, Mode::FloatQuant(k));
   }
 
--- a/pco/src/data_types/float.rs
+++ b/pco/src/data_types/float.rs
@@ -114,10 +114,20 @@ fn choose_mode_and_split_latents<F: Float>(
       let latents = float_mult::split_latents(nums, float_mult_config);
       Ok((mode, latents))
     }
-    ModeSpec::TryFloatQuant(k) => Ok((
-      Mode::FloatQuant(k),
-      float_quant::split_latents(nums, k),
-    )),
+    ModeSpec::TryFloatQuant(k) => {
+      // Must be checked before splitting: split_latents shifts by `k`.
+      if k >= F::L::BITS {
+        return Err(PcoError::invalid_argument(format!(
+          "float quantization k of {} must be less than the type's {} bits",
+          k,
+          F::L::BITS,
+        )));
+      }
+      Ok((
+        Mode::FloatQuant(k),
+        float_quant::split_latents(nums, k),
+      ))
+    }
     ModeSpec::TryIntMult(_) => Err(PcoError::invalid_argument(
       "unable to use int mult mode on floats",
     )),
--- a/pco/src/metadata/delta_encoding.rs
+++ b/pco/src/metadata/delta_encoding.rs
@@ -154,6 +154,16 @@ impl DeltaEncoding {
       2 => {
         let window_n_log = 1 + reader.read_bitlen(BITS_TO_ENCODE_DELTA_LOOKBACK_WINDOW_N_LOG);
         let state_n_log = reader.read_bitlen(BITS_TO_ENCODE_DELTA_LOOKBACK_STATE_N_LOG);
+        // The window buffer is allocated as max(1 << window_n_log, 256) * 2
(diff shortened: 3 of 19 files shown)


mwlon:
Hey thanks, this is great.

pco/src/data_types/float.rs
```diff
@@ -114,10 +114,20 @@ fn choose_mode_and_split_latents<F: Float>(
       let latents = float_mult::split_latents(nums, float_mult_config);
       Ok((mode, latents))
     }
-    ModeSpec::TryFloatQuant(k) => Ok((
-      Mode::FloatQuant(k),
-      float_quant::split_latents(nums, k),
-    )),
+    ModeSpec::TryFloatQuant(k) => {
```
mwlon: I would prefer if we avoid duplicating the validation logic. Instead how about we make each config emit a vec of bids, then we choose the best one, call the generic validation function on it, and then invoke its split_fn?

pco/src/metadata/delta_encoding.rs
```diff
@@ -154,6 +154,16 @@ impl DeltaEncoding {
       2 => {
         let window_n_log = 1 + reader.read_bitlen(BITS_TO_ENCODE_DELTA_LOOKBACK_WINDOW_N_LOG);
         let state_n_log = reader.read_bitlen(BITS_TO_ENCODE_DELTA_LOOKBACK_STATE_N_LOG);
+        // The window buffer is allocated as max(1 << window_n_log, 256) * 2
+        // latents, so an unbounded value here is an allocation the input does
+        // not justify. The encoder never emits more than this.
+        if window_n_log > crate::delta::LOOKBACK_MAX_WINDOW_N_LOG {
```
mwlon: We should have a different constant here to leave room for encoding with longer window_n_logs in the future. I am comfortable imposing a limit here though - I think the log of `MAX_ENTRIES` would make sense, and keep us well under 1GB max allocation here.

pco/src/standalone/decompressor.rs
```diff
@@ -248,7 +248,11 @@ impl FileDecompressor {
   /// analagous file compressor method because the user always knows the dtype
   /// during compression.
   pub fn simple_decompress<T: Number>(&self, mut src: &[u8]) -> PcoResult<Vec<T>> {
-    let mut res = Vec::with_capacity(self.n_hint());
+    // `n_hint` is attacker-controlled and unrelated to how much data is
+    // actually present, so it may only be used as a hint up to a bound the
+    // input can actually justify. Every number costs at least one bit.
+    let cap = std::cmp::min(self.n_hint(), src.len().saturating_mul(8));
```
mwlon: Can you revise/trim the comment a bit and just cap at 2^30? This should be performant by default for outputs of several GB. But I agree terabytes are excessive.

pco/src/mode/float_quant.rs
```diff
@@ -23,10 +23,12 @@ pub(crate) fn join_latents<F: Float>(
   let sign_cutoff = F::L::MID >> k;
   let lowest_k_bits_max = (F::L::ONE << k) - F::L::ONE;
   for ((&y, &m), dst) in primary.iter().zip(secondary.iter()).zip(dst.iter_mut()) {
-    debug_assert!(
-      m >> k == F::L::ZERO,
-      "Invalid input to FloatQuant: m must be a k-bit integer"
-    );
+    // `m` comes off the wire, so this is a corruption check, not an assertion.
+    if m >> k != F::L::ZERO {
+      return Err(PcoError::corruption(
```
mwlon: I think the intent here was really to assert the behavior of our compressor, not decompressor. This check would noticeably slow down decompression, so we should move it to a (debug only) check on the compressor instead.

pco/src/mode/float_quant.rs
mwlon: We could make this a wrapping add.

mwlon:
Can you please split the kani and fuzz changes to separate PRs like the summary suggested?

enomado:
Split done, and all five review comments addressed. This branch is now just the fixes and tests:

* **#383** — the libFuzzer targets and seed corpus
* **#382** — the Kani proofs

The review changes are in a separate commit (`review: validate the mode before splitting; use shared constants`) so the delta since your pass is easy to read.

**Duplicated FloatQuant validation** — done your way, and it turned out better than the check I had: `choose_mode_and_split_latents` is now `choose_mode_bids`, every mode spec emits a `Vec<Bid>`, and `ChunkCompressor::new` picks the winner → `mode_is_valid` → `split_fn`. `mode_is_valid` already rejected `k > PRECISION_BITS`, so the ordering *is* the fix and my extra check is gone. The int and float paths both go through it, and `dict` emits a bid too (it builds its dictionary eagerly and carries the latents in its `split_fn`, since it can't be computed from the mode alone). One knock-on: `Bid`'s bound relaxed from `Number` to `NumberPriv` so the trait method can name `Bid<Self>`.

**Lookback window bound** — now its own constant, `MAX_DELTA_LOOKBACK_WINDOW_N_LOG = MAX_ENTRIES.ilog2() = 24`, in `constants.rs`. `LOOKBACK_MAX_WINDOW_N_LOG` is back to private. That's 256 MiB of window buffer for `i64` at the limit, and leaves 9 doublings of headroom for a future encoder.

**`n_hint`** — capped at `2^30`, comment trimmed. One thing worth flagging: `2^30` is *numbers*, so for `i64` that is still an 8 GiB reservation from a 34-byte file. Fine as a policy decision — it's virtual until touched and you clearly want big outputs to be fast — but it does mean the regression test can't perform the allocation, so it exercises the clamp (`n_hint_prealloc`) instead, plus a live round trip proving absurd hints still decode. Happy to make it a byte budget instead if you'd rather the bound not scale with `size_of::<T>()`.

**FloatQuant `m` check** — agreed, that was an assertion about the compressor. It's a `debug_assert!` in `split_latents` now, and the decoder takes your wrapping add — plus a wrapping sub on `lowest_k_bits_max - m` right above it, which underflows on the same corrupt input for the same reason. Corrupt `m` now decodes to garbage instead of panicking, and the check is off the decode hot path entirely.

Suite is green (134 tests, up 6) and `cargo fmt --check` is clean.

pco/src/wrapped/chunk_compressor.rs
```diff
@@ -490,17 +490,20 @@ impl ChunkCompressor {
     let (mode, latents) = match_number_enum!(
       nums,
       DynNumberSlice<T>(nums) => {
-        let (mode, latents) = T::choose_mode_and_split_latents(nums, config)?;
-        if !T::mode_is_valid(&mode) {
+        let bid = choose_winning_bid(T::choose_mode_bids(nums, config)?);
+        // The mode must be validated before splitting: splitting can use the
```
mwlon: This comment is a bit superfluous

pco/src/standalone/decompressor.rs
```diff
@@ -257,6 +258,14 @@ impl FileDecompressor {
   }
 }
 
+/// How many numbers to preallocate for a file that declares `n_hint`.
+///
+/// `n_hint` is untrusted and unrelated to how much data is actually present,
+/// so we only use it as a hint up to a fixed cap.
+pub(crate) fn n_hint_prealloc(n_hint: usize) -> usize {
```
mwlon: Better to inline this IMO

pco/src/constants.rs
```diff
@@ -37,6 +37,12 @@ pub const MAX_CONSECUTIVE_DELTA_ORDER: usize = 7;
 pub const MAX_CONV1_DELTA_ORDER: usize = 32;
 pub const MAX_CONV1_DELTA_QUANTIZATION: Bitlen = (1 << BITS_TO_ENCODE_DELTA_CONV_QUANTIZATION) - 1;
 pub const MAX_ENTRIES: usize = 1 << 24;
+// The wire format allows a lookback window_n_log of up to 32, but the window
+// buffer is allocated eagerly on read, so we cap what we accept. This leaves
+// room for encoders to grow beyond LOOKBACK_MAX_WINDOW_N_LOG while keeping the
+// buffer well under a GB: no chunk can hold more than MAX_ENTRIES numbers
+// anyway, so a larger window could never be fully used.
+pub const MAX_DELTA_LOOKBACK_WINDOW_N_LOG: Bitlen = MAX_ENTRIES.ilog2() as Bitlen;
```
mwlon, line 45: This name is good, but we should probably change the `delta/mod.rs` constants to clearly be different (e.g. maybe stick "encoding" in there)

pco/src/tests/corruption.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/// `n_hint` is a hint, not a promise: a 26-byte file may legally declare
+/// `usize::MAX` (`standalone/guarantee.rs` writes exactly that), so it must be
+/// clamped before it reaches `Vec::with_capacity`.
+///
+/// This checks the clamp arithmetic rather than decompressing such a file,
+/// since the point of the clamp is that the allocation never happens.
+#[test]
+fn test_absurd_n_hint_is_clamped() {
```
mwlon: This test is a bit silly. If we have something like this it should just assert literal values.

mwlon: If we want to be rigorous, we could add a `standalone::FileDecompressor::with_max_prealloc(bytes: usize)` method (could even make this public), then test the capacity of decompressed vecs.

pco/src/standalone/mod.rs
```diff
@@ -3,7 +3,7 @@ pub use decompressor::{ChunkDecompressor, DecompressorItem, FileDecompressor};
 pub use simple::*;
 
 mod compressor;
-mod constants;
-mod decompressor;
+pub(crate) mod constants;
```
mwlon: Standalone shouldn't leak into the rest of the crate. What are these changes for?

pco/src/metadata/delta_encoding.rs
```diff
@@ -303,6 +312,55 @@ mod tests {
   use super::*;
   use crate::bit_writer::BitWriter;
 
+  // The window buffer is sized `max(1 << window_n_log, 256) * 2` latents, so
```
mwlon: This comment is unnecessary, and we should get rid of the success cases here.

pco/src/metadata/delta_encoding.rs
```diff
@@ -154,6 +154,15 @@ impl DeltaEncoding {
       2 => {
         let window_n_log = 1 + reader.read_bitlen(BITS_TO_ENCODE_DELTA_LOOKBACK_WINDOW_N_LOG);
         let state_n_log = reader.read_bitlen(BITS_TO_ENCODE_DELTA_LOOKBACK_STATE_N_LOG);
+        // The window buffer is allocated as max(1 << window_n_log, 256) * 2
```
mwlon: We should rarely put concrete values into comments. Too easy to grow out of sync with the code.

pco/src/tests/corruption.rs
```diff
@@ -0,0 +1,179 @@
...
+      ChunkConfig::default().with_delta_spec(DeltaSpec::TryLookback),
+    ),
+    (
+      "int_mult",
+      ChunkConfig::default().with_mode_spec(ModeSpec::TryIntMult(7)),
+    ),
+  ]
+}
+
+/// Every single-byte mutation of a valid file must be handled, not crash.
```
mwlon: comment unnecessary

pco/src/tests/corruption.rs
```diff
@@ -0,0 +1,179 @@
...
+        // to garbage -- but reaching this line at all is the assertion.
+        let _ = simple_decompress::<i64>(&corrupt);
+      }
+    }
+  }
+
+  Ok(())
+}
+
+/// Same, for float modes, which have their own latent-splitting arithmetic.
```
mwlon: comment unnecessary

pco/src/tests/corruption.rs
```diff
@@ -0,0 +1,179 @@
...
+/// two saturating values. Cheaper than all 256 values and hits the same
+/// narrow bit fields, since the interesting header fields are 1-5 bits wide.
+fn corrupt_values(orig: u8) -> impl Iterator<Item = u8> {
+  (0..8)
+    .map(move |bit| orig ^ (1 << bit))
+    .chain([0x00, 0xff])
+    .filter(move |&v| v != orig)
+}
+
+fn configs() -> Vec<(&'static str, ChunkConfig)> {
```
mwlon: You just inline the float configs, so I think you should for these too

pco/src/wrapped/chunk_compressor.rs
enomado: Removed.

pco/src/metadata/delta_encoding.rs
enomado: Both done: comment gone, and the test now only walks the rejection cases.

pco/src/metadata/delta_encoding.rs
enomado: Agreed - trimmed to the reason, no sizes. Did the same to the comment on `MAX_DELTA_LOOKBACK_WINDOW_N_LOG` itself.

pco/src/tests/corruption.rs
enomado: Inlined, and dropped the name strings with it - the config `Debug`s fine in the failure message.

pco/src/tests/corruption.rs
enomado: Gone.

pco/src/tests/corruption.rs
enomado: Gone.

pco/src/constants.rs
enomado: Renamed the `delta/mod.rs` pair to `ENCODING_LOOKBACK_MIN_WINDOW_N_LOG` / `ENCODING_LOOKBACK_MAX_WINDOW_N_LOG`, with a one-line note that the reader accepts a looser bound.

pco/src/standalone/mod.rs
enomado: Only so the test could reach `n_hint_prealloc` and `MAX_N_HINT_PREALLOC` - no production reason. Both modules are back to private now; the test goes through the public API instead.

pco/src/standalone/decompressor.rs
enomado: Inlined. It now reads the cap off the `FileDecompressor` too (see the `with_max_prealloc` thread), so it is three lines rather than one:

```rust
let max_prealloc = cmp::min(
  MAX_N_HINT_PREALLOC,
  self.max_prealloc_bytes / mem::size_of::<T>(),
);
let mut res = Vec::with_capacity(cmp::min(self.n_hint(), max_prealloc));
```

pco/src/tests/corruption.rs
enomado: Fair - it was restating the clamp arithmetic back at itself. Went with your next suggestion instead; details in that thread.

enomado: Took the rigorous route. `FileDecompressor::with_max_prealloc(bytes)` now exists, and the test asserts a real capacity:

```rust
let (fd, src) = FileDecompressor::new(file.as_slice())?;
// 8 KiB is 1024 i64s.
let nums = fd.with_max_prealloc(8192).simple_decompress::<i64>(src)?;

assert_eq!(nums, vec![7_i64]);
assert!(nums.capacity() <= 1024, "preallocated {} numbers", nums.capacity());
```

Two calls worth flagging:

* **It is public.** `pub(crate)` plus a single test caller trips `dead_code` in ordinary builds, so public was the honest option. Say the word and I will hide it behind `#[doc(hidden)]` or `#[cfg(test)]`.
* **The unit is bytes, as you wrote it, and it stacks with the fixed cap rather than replacing it:** `min(MAX_N_HINT_PREALLOC, bytes / size_of::<T>())`, defaulting to `usize::MAX` bytes. So out of the box the behaviour is exactly the 2^30 numbers you asked for, and this only ever lowers it.

I also folded away `test_absurd_n_hint_still_round_trips`: the new test round trips the same hostile file, and unlike the old one it does not reserve 8 GiB to do so.

Checked that the test discriminates rather than passing by construction - with the `with_max_prealloc` cap dropped it fails with `preallocated 1073741824 numbers`.

enomado:
All ten comments from the second pass are addressed in `b90a9ec`, again as one commit so the delta is easy to read.

The two that were more than a comment trim:

* **`n_hint` test** - went with your rigorous option: `FileDecompressor::with_max_prealloc(bytes)`, and the test asserts the resulting `Vec` capacity instead of restating the clamp. It stacks with the 2^30 cap rather than replacing it, so the default behaviour is unchanged. Details, and the two judgement calls in it, in that thread.
* **`standalone` leaking** - that was only so the test could reach the helper; both modules are private again and the test goes through the public API.

Also renamed `delta/mod.rs`'s pair to `ENCODING_LOOKBACK_{MIN,MAX}_WINDOW_N_LOG`, dropped the concrete sizes from the comments in both spots, and trimmed the lookback window test down to the rejection cases.

`cargo test -p pco --all-features` is green (133 lib + doc tests), and `clippy --no-deps` reports the same two pre-existing lib warnings as `main`, nothing new.

pco/src/standalone/decompressor.rs
```diff
@@ -248,7 +263,13 @@ impl FileDecompressor {
   /// analagous file compressor method because the user always knows the dtype
   /// during compression.
   pub fn simple_decompress<T: Number>(&self, mut src: &[u8]) -> PcoResult<Vec<T>> {
-    let mut res = Vec::with_capacity(self.n_hint());
+    // `n_hint` is untrusted and unrelated to how much data is actually
+    // present, so we only follow it up to a cap.
+    let max_prealloc = cmp::min(
+      MAX_N_HINT_PREALLOC,
```
mwlon: Since we have max_prealloc_bytes now let's get rid of this constant and make the default max_prealloc_bytes something like 2^32

enomado: Done in `bdb9f7e`. `MAX_N_HINT_PREALLOC` is gone, so the byte budget is the only bound now, defaulting to `DEFAULT_MAX_PREALLOC_BYTES = 2^32`.

One wrinkle: `1 << 32` does not fit a 32-bit `usize`, and `bit_reader.rs` mentions wasm, so the constant saturates rather than being a plain literal:

```rust
pub const DEFAULT_MAX_PREALLOC_BYTES: usize = if (1_u64 << 32) <= usize::MAX as u64 {
  (1_u64 << 32) as usize
} else {
  usize::MAX
};
```

Happy to drop the branch for a bare `1 << 32` if you do not care about 32-bit targets.

Net effect at the default: 2^29 `i64`s instead of the old 2^30, and no more scaling of the cap with `size_of::<T>()` in the wrong direction. `cargo test -p pco --all-features` is green (133 lib + 5 doc), `fmt --check` clean, clippy reports only the two pre-existing lib warnings.

mwlon:
Tests failed due to unrelated linting updates. Merge with main and try again

enomado:
Merged `main` in (`f53bf47`). Confirmed that was it: the failing job was `ruff check`, and all four hits (`generate_randoms.py` DTZ004, two I001, `test_wrapped.py` RUF059) are in files this branch never touches - your `ruff fixes` commit covers all of them. `ruff 0.16.1` on the merged tree passes clean, and `cargo test -p pco --all-features` is still green (133 lib + 5 doc).

mwlon (approved):

PR #381, opened 2026-07-31
Harden the C API: two defects and a soundness fix

enomado:
Rebased onto main (1.0.3) and cut down to **one commit that touches only the C API**, so nothing here overlaps what #380 already merged. The harnesses that found these moved out to #383; this PR is fixes only.

The C API had no harness at all. That is where memory safety stops being Rust's problem: three exported functions, every one of them writing through a pointer the caller owns, and the only thing exercising them was one happy-path C file (six f64s, default config). Two guard-banded libFuzzer targets (#383) found no out-of-bounds write. What they did find:

## 6. `guarantee_file_size` ignored the config it was sizing for

The header documents a sequence — ask `pco_standalone_guarantee_file_size` how big the output can get, allocate that, then `pco_standalone_simple_compress_into`. But the guarantee took only `(n, dtype)` and computed with `PagingSpec::default()`, while compression honoured the caller's `max_page_n`: one chunk per page, each with its own overhead.

21 i64s with `max_page_n = 1`: the bound says 340 bytes, the file is 368. A caller following the documented sequence exactly gets `PcoCompressionError`, and there is no API that would have told them the right size.

Fixed additively — new `pco_standalone_guarantee_file_size_with_config` — rather than by adding a parameter to the existing symbol, which would break compiled consumers silently. The old function now documents itself as assuming the default spec.

## 7. An empty input skipped config validation entirely

`simple_compress(&[], config)` returned `Ok` for a config the same function rejects with one element in it (`compression_level = 1_577_058_303`, say). Validation lives in `ChunkCompressor::new`, and an empty input produces no chunks, so nothing ever looked at the config — while the doc comment says "will return an error if the compressor config is invalid".

This one is `pco`, not `pco_c`; a Rust caller sees it too. It matters most through the C API, where compressing an empty array is the natural way to ask "is this config OK?" and the answer was a false yes.

Fixed by validating once up front in `simple_compress_into` and `simple_compress_dyn`.

## 8. `pco_standalone_simple_decompress_into` was a safe `fn`

It dereferences four caller-supplied pointers and was declared `pub extern "C" fn`, unlike its compression twin two functions above it. A safe signature claims no arguments can make the function misbehave, which is exactly false here. Rust-side only — the C ABI and the generated header are unchanged.

## Also

`pco_c` gains `rlib` in `crate-type`. Without it the crate has no target any Rust code can link, so nothing in-process — test or fuzz harness — could call the C API at all. Costs a C consumer nothing, and #383's two C API targets need it.

## Verification

* `cargo test -p pco -p cpcodec --all-features`: 134 + 3 + 5 doctests, green
* `pco_c/test/run_test.sh` covers finding 6 from the C side, and fails loudly if the paged file does *not* outgrow the default-spec bound — a regression test for a size bound proves nothing if the scene never exceeds it
* `rustfmt --check` clean; `cargo clippy --no-deps --all-features` introduces no new warnings (the pre-existing one in `errors.rs:59` is left alone)
* Post-fix fuzzing, no crash: `c_api_roundtrip` 617 924 runs / 201 s, `c_api_decompress` 690 292 runs / 201 s

Full write-up in `fuzz/FINDINGS.md` (#383).

## Known, not fixed

* `_decompress_into` decodes into a `Vec` and only then compares the count against `dst_cap`, so a caller with a 10-element destination can still be made to materialise the whole file. Bounded by the input's real size now that #380 is in, but the shape is "decode everything, then check". Happy to change it if you want the count checked against the chunk header first.
* An out-of-range `compression_level` comes back as `PcoCompressionError`, indistinguishable from a genuine compression failure. There is a `PcoInvalidType` for a bad dtype but no equivalent for a bad config; adding one is an ABI question I did not want to decide unilaterally.
* `pco_python` and `pco_java` are still untouched. Neither can be driven from a Rust fuzz target without standing up the respective runtime.


mwlon:
Let me know if you wanted to clean this one up or not. It's hard for me to tell right now how much of this (if any) is additive with the other fixes we already merged in. The C-API-only changes are probably good.

enomado:
Cleaned it up: rebased onto main and cut to one commit that touches only `pco_c` plus the one `pco` fix the C API surfaced (empty input skipping config validation). Nothing here overlaps #380 — you can read the whole diff top to bottom now.

The two guard-banded fuzz targets that found these moved to #383, so the split you asked for earlier holds: fixes here, harnesses there. One coupling worth knowing: those targets link `cpcodec` as a Rust dependency, so they need the `rlib` crate-type that this PR adds. Merging this one first keeps #383 building.

pco_c/include/cpcodec_generated.h
```diff
@@ -25,13 +25,37 @@ typedef struct PcoChunkConfig {
...
+ *
+ * If you pass a `PcoChunkConfig` with a non-zero `max_page_n` to
+ * `pco_standalone_simple_compress_into`, use
+ * `pco_standalone_guarantee_file_size_with_config` instead: the file is split
+ * into one chunk per page, each with its own overhead, so a smaller
+ * `max_page_n` means a *larger* upper bound than this function reports.
  *
  * This function is thread-safe and performs no heap allocation.
  */
 size_t pco_standalone_guarantee_file_size(size_t n, unsigned char dtype);
```
mwlon, line 39: Just break the API and add chunk config as an argument here rather than making multiple functions.

pco_c/src/lib.rs
```diff
@@ -23,8 +24,11 @@ pub enum PcoError {
 /// Only `compression_level` and `paging_spec` are supported for now; other
 /// fields can be added later without breaking the ABI.
 #[repr(C)]
+// Not `Copy`: it would make `to_chunk_config(&self)` a clippy `wrong_self_convention`
```
mwlon, line 27: unnecessary comment

pco_c/src/lib.rs
```diff
@@ -120,8 +133,14 @@ fn _decompress_into<T: Number>(
 }
 
 /// Return the maximum possible byte size of a standalone file for `n`
-/// elements of `dtype`.  Returns 0 for an invalid `dtype` or invalid
-/// paging spec.
+/// elements of `dtype`, **assuming the default paging spec**.  Returns 0 for
+/// an invalid `dtype` or invalid paging spec.
+///
+/// If you pass a `PcoChunkConfig` with a non-zero `max_page_n` to
```
mwlon, line 139: Too many clop comments; most of this is readable from the code, and what isn't should be commented in exactly one place.

pco/src/standalone/simple.rs
```diff
@@ -60,6 +69,14 @@ pub fn simple_compress<T: Number>(src: &[T], config: &ChunkConfig) -> PcoResult<
 }
 
 pub fn simple_compress_dyn(src: DynNumberSlice, config: &ChunkConfig) -> PcoResult<Vec<u8>> {
+  // Same reason as in `simple_compress_into`: no chunks, hence no per-chunk
+  // validation, hence an empty input used to accept an invalid config.
+  let latent_type = match_number_enum!(
```
mwlon, line 74: Ah, this function was apparently exposed by mistake in 1.0.1. It shouldn't be `pub`, and hence you can validate in `simple_compress`.

PR #382, opened 2026-08-01
Kani proof harnesses for the unsafe decode path

enomado:
Split out of #380 as requested.

14 Kani proofs over the unsafe decode path, all verifying:

* the padding contract for all three `read_uint_at` arms,
* `u64_at` / `u32_at` bounds,
* `lowest_bits` and `lowest_bits_fast`,
* `calc_max_bytes` against the dispatch it feeds,
* the ANS index arithmetic,
* the unchecked window-buffer indexing in `delta/lookback.rs`.

The ANS one doubles as a regression gate: raise `MAX_ANS_BITS` and `stride * step` overflows `u32`, and the proof goes red. A negative control confirmed the lookback harness has teeth.

Everything is behind `#[cfg(kani)]`, so it is out of the normal build; the only non-gated change is a `check-cfg` lint entry in `pco/Cargo.toml` so `cfg(kani)` doesn't warn.

Run with `cargo kani -p pco`.

Companion PRs: #380 (the fixes these and the fuzzing found) and the fuzzing harness.

### DIFF
--- a/pco/src/kani_proofs.rs
+++ b/pco/src/kani_proofs.rs
@@ -1,0 +1,263 @@
+//! Kani proof harnesses for the unsafe decode path.
+//!
+//! The crate reads compressed data through raw-pointer loads in
+//! `bit_reader` that have no bounds checks. Their safety rests on a
+//! *padding contract*: the caller must guarantee the source slice has
+//! enough bytes past the current index. These harnesses state that
+//! contract explicitly and let Kani check it for every input, rather
+//! than for the fixed corpus the unit tests use.
+
+use crate::bit_reader::{read_uint_at, u32_at, u64_at, BitReader};
+use crate::bits;
+use crate::constants::{Bitlen, Weight, MAX_ANS_BITS};
+use crate::read_write_uint::calc_max_bytes;
+
+// Enough room that a nondeterministic byte_idx still has interesting
+// slack, but small enough to keep the model tractable.
+const SRC_LEN: usize = 32;
+
+// ---------------------------------------------------------------------------
+// bits::lowest_bits{,_fast}
+// ---------------------------------------------------------------------------
+
+// `lowest_bits_fast` documents "doesn't handle the case when n >= U::BITS".
+// Below that bound it must be a plain mask, with no shift overflow.
+#[kani::proof]
+fn proof_lowest_bits_fast_u64_is_a_mask() {
+  let x: u64 = kani::any();
+  let n: Bitlen = kani::any();
+  kani::assume(n < 64);
+
+  let got = bits::lowest_bits_fast(x, n);
+  let want = x & ((1u64 << n) - 1);
+  assert_eq!(got, want);
+}
+
+#[kani::proof]
+fn proof_lowest_bits_fast_u32_is_a_mask() {
+  let x: u32 = kani::any();
+  let n: Bitlen = kani::any();
+  kani::assume(n < 32);
+
+  let got = bits::lowest_bits_fast(x, n);
+  let want = x & ((1u32 << n) - 1);
+  assert_eq!(got, want);
+}
+
+// The checked variant must be total: no n at all may panic, and n >= BITS
+// must be the identity.
+#[kani::proof]
+fn proof_lowest_bits_u64_is_total() {
+  let x: u64 = kani::any();
+  let n: Bitlen = kani::any();
+
+  let got = bits::lowest_bits(x, n);
+  if n >= 64 {
+    assert_eq!(got, x);
+  } else {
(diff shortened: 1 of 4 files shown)


enomado:
Heads up on the red `Python` check: it is not from this branch. The job is `ruff check .`, and all four findings are in files this PR does not touch (`pco_cli/generate_randoms.py`, `pco_python/docs/conf.py`, `pco_python/test/test_standalone.py`, `pco_python/test/test_wrapped.py`) - this branch adds no Python at all. The workflow uses `jpetrucciani/ruff-check@main`, so it floats to whatever ruff is current (0.16.1 today) and picked up rules that did not exist when `main` last ran green in July: `DTZ004`, two `I001`, and one `RUF059`.

`Rust` and `Java` are green here. Happy to send a separate one-commit PR for those four if you want CI back to green.

pco/src/kani_proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+// slack, but small enough to keep the model tractable.
+const SRC_LEN: usize = 32;
+
+// ---------------------------------------------------------------------------
+// bits::lowest_bits{,_fast}
+// ---------------------------------------------------------------------------
+
+// `lowest_bits_fast` documents "doesn't handle the case when n >= U::BITS".
+// Below that bound it must be a plain mask, with no shift overflow.
+#[kani::proof]
```
mwlon, line 25: Should this file exist? Why not put these into the respective files for the things they test?

pco/src/kani_proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+  let n: Bitlen = kani::any();
+  kani::assume(n < 32);
+
+  let got = bits::lowest_bits_fast(x, n);
+  let want = x & ((1u32 << n) - 1);
+  assert_eq!(got, want);
+}
+
+// The checked variant must be total: no n at all may panic, and n >= BITS
+// must be the identity.
```
mwlon, line 48: This seems pretty silly - there's nothing unsafe in the function it tests, and it essentially repeats the logic of the function.

pco/src/kani_proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+// The comment in bit_reader.rs justifies the three arms by hand:
+//   0..=25  bits -> 4 bytes,  26..=57 bits -> 8 bytes,  58..=113 bits -> 15.
+// `read_uint` picks the arm from `U::MAX_BYTES`, which for the types this
+// crate actually instantiates works out to:
+//   u8, u16 -> 4 bytes    u32 -> 8 bytes    u64, usize -> 15 bytes
+// Each harness pins one arm: given `bits_past_byte < 8` (what `refill`
+// establishes) and n within that arm's stated range, the read touches only
+// READ_BYTES bytes past byte_idx and no shift overflows.
+
+#[kani::proof]
```
mwlon, line 111: I'm kinda skeptical of the value of these tests. It's apparent enough to me that there's an extremal, hardest example to test for each of these, which could simply be a unit test.

pco/src/kani_proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// calc_max_bytes: does the advertised padding actually cover the reads?
+// ---------------------------------------------------------------------------
+
+// `calc_max_bytes(precision)` is what callers use to size padding, and
+// `read_uint` picks its READ_BYTES arm from `U::MAX_BYTES`. These must agree:
+// the arm chosen for a type must never read past that type's MAX_BYTES.
+#[kani::proof]
```
mwlon, line 164: again doesn't seem like this one tests anything unsafe

pco/src/kani_proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+// the loop-free facts it depends on, for every admissible size_log:
+//
+//   * the shift in `mod_table_size` does not underflow,
+//   * `stride * step` does not overflow Weight,
+//   * the resulting index is always < table_size, so `res[..]` can't panic.
+//
+// This is also a regression gate: if MAX_ANS_BITS is ever raised, the
+// multiplication overflows and this harness starts failing.
+#[kani::proof]
+fn proof_ans_spread_index_arithmetic() {
```
mwlon, line 224: This test seems nearly trivial

mwlon:
You can merge main to fix the spurious failures

enomado:
This is mostly slop. I just ask ai to do something and it goes. I fix it in week, when i get unutilised tokens

mwlon:
I think the value of this particular PR is minimal, so don't worry about de-slop-ifying it. Just gonna close.

PR #383, opened 2026-08-01
libFuzzer targets for the decode, roundtrip and C API paths

enomado:
Split out of #380 as requested; rebased onto main (1.0.3) and now carries the C API harnesses too, since #381 was cut down to fixes only.

Five `cargo fuzz` targets plus a minimized seed corpus:

* **`decompress_corrupt`** — the productive one. It compresses real data first and *then* patches bytes, so mutations land on parsed metadata instead of bouncing off the magic header.
* **`decompress_arbitrary`** — unstructured input straight at the decoder.
* **`roundtrip`** — compress/decompress equivalence across generated configs.
* **`c_api_roundtrip`** / **`c_api_decompress`** — the FFI surface, with 64-byte `0xA5` guard bands either side of every caller-owned buffer; without a sanitizer that is what makes a write past the end observable rather than landing in allocator slack. Same trick as `decompress_corrupt`: compress real numbers through the C API first, then corrupt — 202 coverage edges before that change, 3897 after.

⚠️ The two C API targets link `cpcodec` as a Rust dependency, which needs the `rlib` crate-type added in #381. Merge that one first, or these two targets will not build (the other three are unaffected).

`fuzz/` declares its own `[workspace]`, so `cargo test` and `cargo build` at the root are unaffected; nothing here is part of the normal build.

Run with e.g. `cargo +nightly fuzz run decompress_corrupt`. Seeds unpack from `fuzz/seeds/*.tar.zst` (see `fuzz/seeds/README.md`); `fuzz/corpus/` stays gitignored as cargo-fuzz expects.

Note that `cargo fuzz` builds at `opt-level=3` **with** debug assertions and overflow checks, which is not what ships — worth re-checking any finding in stock release too, since the two modes disagree.

`fuzz/FINDINGS.md` is the write-up of what these turned up (eight defects) and how each one is reached. Findings 1–5 are fixed by #380, already merged and released in 1.0.3; findings 6–8 by #381. Line numbers there refer to the 1.0.2 tree, before the fixes.


enomado:
Heads up on the red `Python` check: it is not from this branch. The job is `ruff check .`, and all four findings are in files this PR does not touch (`pco_cli/generate_randoms.py`, `pco_python/docs/conf.py`, `pco_python/test/test_standalone.py`, `pco_python/test/test_wrapped.py`) - this branch adds no Python at all. The workflow uses `jpetrucciani/ruff-check@main`, so it floats to whatever ruff is current (0.16.1 today) and picked up rules that did not exist when `main` last ran green in July: `DTZ004`, two `I001`, and one `RUF059`.

`Rust` and `Java` are green here. Happy to send a separate one-commit PR for those four if you want CI back to green.

fuzz/fuzz_targets/decompress_arbitrary.rs
```diff
@@ -0,0 +1,30 @@
...
+use libfuzzer_sys::fuzz_target;
+
+fn try_all(src: &[u8]) {
+  // Each of these instantiates a different latent width, and therefore a
+  // different arm of `read_uint_at`.
+  let _ = pco::standalone::simple_decompress::<u8>(src);
+  let _ = pco::standalone::simple_decompress::<u16>(src);
+  let _ = pco::standalone::simple_decompress::<u32>(src);
+  let _ = pco::standalone::simple_decompress::<u64>(src);
+  let _ = pco::standalone::simple_decompress::<i16>(src);
```
mwlon, line 20: Include i8 just for completion?

fuzz/fuzz_targets/decompress_corrupt.rs
```diff
@@ -0,0 +1,72 @@
...
+//! Contract: any corrupted stream must produce an error, not a panic, not an
+//! out-of-bounds read, and not an unbounded allocation.
+
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+use pco::{ChunkConfig, DeltaSpec, ModeSpec, PagingSpec};
+
+#[derive(Arbitrary, Debug)]
+struct Patch {
+  offset: u16,
```
mwlon, line 19: Should this just be a usize?

fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -0,0 +1,122 @@
...
+      want
+    );
+  }
+}
+
+fuzz_target!(|input: Input| {
+  let config = config(&input);
+  let raw = &input.nums;
+  match input.dtype % 10 {
+    0 => check::<u8, _>(&raw.iter().map(|&x| x as u8).collect::<Vec<_>>(), &config, |&x| x as u64),
```
mwlon, line 97: I think you could just pass input.nums here without referencing raw

fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -0,0 +1,122 @@
...
+      got,
+      want
+    );
+  }
+}
+
+fuzz_target!(|input: Input| {
+  let config = config(&input);
+  let raw = &input.nums;
+  match input.dtype % 10 {
```
mwlon, line 96: I think you should build this and the other match arms using `NumberType` and `match_number_enum`.

fuzz/FINDINGS.md
```diff
@@ -0,0 +1,113 @@
+# Findings
```
mwlon, line 1: This doc is already out of date; just delete

fuzz/seeds/README.md
```diff
@@ -0,0 +1,56 @@
+# Seed corpora
```
mwlon, line 1: This should be trimmed down to only instructions.

mwlon: Also it should be in fuzz/, not seeds/

PR #386, opened 2026-08-03
Variable sampling ratio

mwlon:
* Build larger samples (for auto mode and delta encoding) at higher compression levels, smaller samples at lower levels.
  * Motivation is to increase the strength  of the compression level tradeoff
* Use fewer, larger contiguous groups for delta sample.
  * Motivation is that lookback delta encoding is really baffled by jumps between groups, so we don't choose it as often as we could
  * This change makes the default group ~400 instead of 200, which is a slightly help, nothing crazy
* Change mode sampling algorithm to guarantee intended sample size via Floyd's algorithm
  * This is also faster than the "persistence" approach I had before
* Change float mult heuristics to always consider both GCD and trailing zeros approaches
  * Motivation is that I noticed sometimes on the taxi dataset, we choose by trailing zeros and overlook the superior result from the GCD heuristics
  * This adds a small amount of compute, but empirically it's cheap. It only affects cases with a lot of floats with many trailing zeros (like the taxi dataset tips column)

2M rows of taxi dataset before and after:
```
│ <sum>                    │ pco:level=0  │ 152.008452ms │           0ns │        86691650 │
│ <sum>                    │ pco:level=1  │ 300.737578ms │           0ns │        50056289 │
│ <sum>                    │ pco:level=2  │ 328.981058ms │           0ns │        41733496 │
│ <sum>                    │ pco:level=3  │ 351.580873ms │           0ns │        39469055 │
│ <sum>                    │ pco:level=4  │ 369.008725ms │           0ns │        37626435 │
│ <sum>                    │ pco:level=5  │ 387.142225ms │           0ns │        36337150 │
│ <sum>                    │ pco:level=6  │ 405.473831ms │           0ns │        35324805 │
│ <sum>                    │ pco:level=7  │ 427.394457ms │           0ns │        34856536 │
│ <sum>                    │ pco          │ 454.924829ms │           0ns │        34604045 │
│ <sum>                    │ pco:level=9  │ 509.401017ms │           0ns │        34477461 │
│ <sum>                    │ pco:level=10 │ 668.895165ms │           0ns │        34314969 │
│ <sum>                    │ pco:level=11 │ 1.141470558s │           0ns │        34213894 │
│ <sum>                    │ pco:level=12 │ 2.228467933s │           0ns │        34181389 │

│ <sum>                    │ pco:level=0  │ 122.307682ms │           0ns │        86004182 │
│ <sum>                    │ pco:level=1  │  275.62887ms │           0ns │        49964474 │
│ <sum>                    │ pco:level=2  │ 305.476309ms │           0ns │        41704445 │
│ <sum>                    │ pco:level=3  │ 330.863079ms │           0ns │        39468205 │
│ <sum>                    │ pco:level=4  │ 362.607516ms │           0ns │        37483697 │
│ <sum>                    │ pco:level=5  │ 377.895269ms │           0ns │        36363017 │
│ <sum>                    │ pco:level=6  │ 402.457099ms │           0ns │        35324805 │
│ <sum>                    │ pco:level=7  │ 427.939246ms │           0ns │        34856536 │
│ <sum>                    │ pco          │ 464.195309ms │           0ns │        34603523 │
│ <sum>                    │ pco:level=9  │ 528.474683ms │           0ns │        34477608 │
│ <sum>                    │ pco:level=10 │ 691.465954ms │           0ns │        34314969 │
│ <sum>                    │ pco:level=11 │ 1.187880183s │           0ns │        34213894 │
│ <sum>                    │ pco:level=12 │ 2.567288498s │           0ns │        34181389 │
```


pco/src/data_types/float.rs
```diff
@@ -459,7 +463,7 @@ mod tests {
   #[test]
   fn test_choose_mult_mode() {
     let base = 1.5;
-    let nums = (0..1000).map(|i| (i as f64) * base).collect::<Vec<_>>();
```
mwlon, line 462: I did a bunch of digging, and it was just lucky that 1000 was enough to infer the best base before. The sample size goes from 34 -> 59 as we double the nums, and when I look at the stats from the 34 nums under the new sampling, they look much more often like multiples of 3.0 than 1.5.

Skielex (commented):
Code-wise it looks fine.

The question is if we should do a thorough cross-platform performance review?

mwlon:
@Skielex Yep, I realized after submitting this the compression ratio changes aren't super straightforward on all datasets (but I expect performance to look similar everywhere). I might put this on hold and make a different change first.

mwlon:
Going to break this into smaller pieces. Some should be obvious, others require more work

PR #406, opened 2026-09-20
Codspeed CI

mwlon:
* Run on bare metal codspeed ARM instance.
* Triggered for every main commit, manual run through codspeed, and PR commit when the `bench` label is applied.
  * Only `triage` members of the repo have permissions to apply the bench label.


mwlon:
!bench

